PR #3, opened 2020-01-05
Added OLED black theme.

johanw666:
Added OLED black theme.


valldrac:
Thanks for the PR. I'll review it after the first version is released.

Meteor0id:
It would be nicer if the existing dark theme was modified to be OLED compatible, but of course that might be even more maintenance.

nutts0:
Can u merg this

valldrac:
I chose to replace the current dark theme instead of creating a new theme. It would be a mess to mantain it.

PR #4, opened 2020-01-18
Check android sdk/ndk sha256 sum on Dockerfile

spk:
Hi Oscar, thanks for the ping I've applied the patch updated from Signal, cheers, Laurent


valldrac:
Merged! Thanks a lot.

PR #5, opened 2020-02-26
Added an option to not show v2 PIN reminders.

johanw666:
This extra option is shown when a v2 PIN set and it prevents the weekly reminder to show.

One small issue: if a reminder is already shown you have to exit Molly (locking is not necessary, just exit the main activity) and reopen it for the reminder to go away. I did not fix this because it would complicate the code a lot for an option that is probably set-once and forget.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
@@ -226,9 +226,7 @@ public class ConversationListFragment extends MainFragment implements LoaderMana
     initializeTypingObserver();
     initializeSearchListener();
 
-    if (!TextSecurePreferences.isPinV2ReminderDisabled(requireContext())) {
-      RegistrationLockDialog.showReminderIfNecessary(this);
-    }
+    RegistrationLockDialog.showReminderIfNecessary(this);
 
     TooltipCompat.setTooltipText(searchAction, getText(R.string.SearchToolbar_search_for_conversations_contacts_and_messages));
   }
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
@@ -226,7 +226,9 @@ public class ConversationListFragment extends MainFragment implements LoaderMana
     initializeTypingObserver();
     initializeSearchListener();
 
-    RegistrationLockDialog.showReminderIfNecessary(this);
+    if (!TextSecurePreferences.isPinV2ReminderDisabled(requireContext())) {
+      RegistrationLockDialog.showReminderIfNecessary(this);
+    }
 
     TooltipCompat.setTooltipText(searchAction, getText(R.string.SearchToolbar_search_for_conversations_contacts_and_messages));
   }
--- a/app/src/main/java/org/thoughtcrime/securesms/megaphone/Megaphones.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/megaphone/Megaphones.java
@@ -20,6 +20,7 @@ import org.thoughtcrime.securesms.lock.v2.KbsMigrationActivity;
 import org.thoughtcrime.securesms.lock.v2.PinUtil;
 import org.thoughtcrime.securesms.logging.Log;
 import org.thoughtcrime.securesms.util.FeatureFlags;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
 
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/thoughtcrime/securesms/util/TextSecurePreferences.java
```diff
@@ -187,6 +187,16 @@
 
   private static final String STORAGE_MANIFEST_VERSION = "pref_storage_manifest_version";
 
+  public static final String DISABLE_PINV2_MEMINDERS = "pref_signal_disable_pinv2_reminders";
```
valldrac: Typo *DISABLE_PINV2_REMINDERS

app/src/main/res/values/strings.xml
```diff
@@ -1874,6 +1874,8 @@
     <string name="preferences_app_protection__change">Change</string>
     <string name="preferences_app_protection__create">Create</string>
     <string name="preferences_app_protection__your_pin_adds_an_extra_layer_of_security_and_backs">Your PIN adds an extra layer of security and backs up your account. You\'ll be asked for it when you register your phone number with Signal. If you forget your PIN, you\'ll be locked out of your account for 7 days.</string>
+    <string name="preferences_app_protection__disable_pinv2_reminders">Disable PIN reminders</string>
+    <string name="preferences_app_protection__disable_pinv2_reminders_summary">Disable PIN reminder requests</string>
```
valldrac: The summary could be more explicit like "Do not prompt to enter your PIN periodically"

app/src/main/java/org/thoughtcrime/securesms/megaphone/Megaphones.java
```diff
@@ -78,11 +79,13 @@ private Megaphones() {}
...
    */
-  private static Map<Event, MegaphoneSchedule> buildDisplayOrder() {
+  private static Map<Event, MegaphoneSchedule> buildDisplayOrder(Context context) {
     return new LinkedHashMap<Event, MegaphoneSchedule>() {{
       put(Event.REACTIONS, new ForeverSchedule(true));
       put(Event.PINS_FOR_ALL, new PinsForAllSchedule());
-      put(Event.PIN_REMINDER, new SignalPinReminderSchedule());
+      if (!TextSecurePreferences.isPinV2ReminderDisabled(context)) {
+        put(Event.PIN_REMINDER, new SignalPinReminderSchedule());
+      }
```
valldrac: An easy way to fix the problem you mentioned in the PR is to modify the class `SignalPinReminderSchedule` instead. Add something like this to the method `shouldDisplay`:
```
if (TextSecurePreferences.isPinV2ReminderDisabled(ApplicationDependencies.getApplication())) {
  return false;
}
```

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
```diff
@@ -226,7 +226,9 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
     initializeTypingObserver();
     initializeSearchListener();
 
-    RegistrationLockDialog.showReminderIfNecessary(this);
+    if (!TextSecurePreferences.isPinV2ReminderDisabled(requireContext())) {
+      RegistrationLockDialog.showReminderIfNecessary(this);
+    }
```
valldrac: Better to do this check inside the body of the method `showReminderIfNecessary`

valldrac:
Thanks for the PR. Please review my comments.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
johanw666: That is a static method so I'd have to pass the context there. It can be done but will require more maintenance in the future.

app/src/main/java/org/thoughtcrime/securesms/util/TextSecurePreferences.java
johanw666: Oops... I'll change that.

app/src/main/res/values/strings.xml
johanw666: OK.

app/src/main/java/org/thoughtcrime/securesms/megaphone/Megaphones.java
johanw666: I didn't see this way to get the context to this class myself, I'll test this approach.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
johanw666: OK, if you like. I just copied the method OWS uses in onResume() for the sms reminder.

PR #43, opened 2021-04-30
Remove arbitary 3 hour limit for sending Remote Delete Requests

dylangerdaly:
Allow for 1 day to send a Remote Deletion Request, sets the `SEND_THRESHOLD` to be the same as the `RECEIVE_THRESHOLD`.

I'm not sure why there are Deletion Thresholds in the first place.

TL;DR gives you 1 day to redact/delete messages as apposed to the stock 3 hours.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/util/RemoteDeleteUtil.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/util/RemoteDeleteUtil.java
@@ -13,7 +13,7 @@ import java.util.concurrent.TimeUnit;
 public final class RemoteDeleteUtil {
 
   private static final long RECEIVE_THRESHOLD = TimeUnit.DAYS.toMillis(1);
-  private static final long SEND_THRESHOLD    = TimeUnit.HOURS.toMillis(3);
+  private static final long SEND_THRESHOLD    = TimeUnit.DAYS.toMillis(1);
 
   private RemoteDeleteUtil() {}
 


benarmstead:
I like the idea of this. 
Have you tested it to ensure that there is no receiver side validation for whether the sender has deleted it within 3 hours or not?
If so then I think this is a good change, maybe even disable the timer for deleting messages entirely?
Would be interesting to see what @valldrac thinks of this.

dylangerdaly:
Yeah I've tested it successfully, legitimately used it a few times on my own build too.

In terms of receiver side validation, that's `RECEIVE_THRESHOLD`, currently set to 1 day (set by Signal), I'm simply matching `SEND_THRESHOLD` from 3 hours, to the receiver side validation of 1 day

benarmstead:
Cool, sounds good to me then, its a feature I would like to see. 
I don't really like the time limit at all but if as you say there is a 1 day `RECEIVE_THRESHOLD`, then it seems we are stuck with 1 day to maintain compatibility.

Would be interesting to know what happened if you turned your networking off, and deleted a message at e.g. 23 hours 59 mins after being sent, then at 24 hours and 1 min turned your networking back on. I'm assuming the error wouldn't be too bad though as it would still be possible with the official signal client given the deleter being offline long enough.

benarmstead (approved):
The change is good, but testing needs to be done by the terms of my latest comment here #43 to ensure the error thrown is not too bad.

valldrac:
This was discussed in the IRC room days ago and I over looked it.

In summary, there are social engineering issues with longer time frames for delete requests. In my opinion, we shouldn't "abuse" this feature to have a way to delete content of remote devices (a legitimate use case) because in my opinion that would be require explicit previous approval from the receiver.

Still the current feature is useful for the common wrong-chat mistakes and alike.

The shared knowledge is that Signal allows up to 3 hours limit. The official documentation also states it. But as we see, a rogue client would be able to delete content within 24 hours, because the hard limit is enforce in the receiver side. My guess is that Signal did this way because the timestamps are set when the messages enter the server's queue, and thus they allow the sender to delete messages while it is offline. They probably made this to match user expectations.

But I don't want Molly users to discover that 3 hours was not the hard limit. So I propose to don't take that compromise, and sets both timers to 3 hours instead.

dylangerdaly:
Personally I disagree, but it's your project, you have reserve powers.

I think that Signal shouldn't dictate what forks can and can't do, giving Molly users more power over their own sent messages I think is a plus, I think Molly **and** Signal should have the redaction receiver check totally removed, sent/user generated content should be withdraw-able, at anytime, just as consent is.

I actually used this a few days ago, I had sent a friend a long message, and I was able to redact 90% of it within the 5th hour of sending it (Timezones), it simply gives the user more control over their own content.

However I can understand not wanting to 'rock the boat' in terms of changing what Signal is expecting.

PR #56, opened 2021-06-29
Minor grammar correction in strings2.xml

taivlam:
Minor grammar gerund correction (change "remain" to "remaining" in "Clearing unencrypted data remain in RAM" message)

### DIFF
--- a/app/src/main/res/values/strings2.xml
+++ b/app/src/main/res/values/strings2.xml
@@ -31,7 +31,7 @@
     <string name="preferences_network__online">Online</string>
     <string name="ApplicationPreferencesActivity_privacy_summary_passphrase_registration_locks">Database encryption %1$s, Registration lock %2$s</string>
     <string name="WipeMemoryService_secure_wipe_in_progress">Secure wipe in progress</string>
-    <string name="WipeMemoryService_clearing_unencrypted_data_remain_in_ram">Clearing unencrypted data remaining in RAM</string>
+    <string name="WipeMemoryService_clearing_unencrypted_data_remain_in_ram">Clearing unencrypted data in RAM</string>
     <string name="PrivacySettingsFragment_data_at_rest">Data at rest</string>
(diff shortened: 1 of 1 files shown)


valldrac:
Thanks for bringing this up. It's a grammar error indeed.

We discussed this in the Matrix room and they suggested to better replace this string by 'Clearing unencrypted data from RAM". It's shorter and avoid the use of two gerunds in the same phrase.

Would you like to do this modification?

Note that the identifier `WipeMemoryService_clearing_unencrypted_data_remain_in_ram` should be changed accordingly too, both in the XML and in the source code (I think it's only referenced in one Java file). This is needed to sync the translations.

taivlam:
@valldrac Yes, I think the shorter string, "Clearing unencrypted data from RAM" is better.

Initially, I was hesitant to modify the XML identifier because I didn't know how many times I had to replace that everywhere in the repository, since there are quite a few other translations built in Molly.

I'll change the identifier here accordingly (if I can figure how to modify this pull request on GitHub) and I believe I've located the mentioned [Java file](https://github.com/mollyim/mollyim-android/blob/3c61b6c15367ced4fb2a5b66d3cd1467805144cd/app/src/main/java/org/thoughtcrime/securesms/service/WipeMemoryService.java).

valldrac:
Great. Please, do not replace the identifier in the translations. Only in the main language (English) XML file and in the Java code. The Weblate service will take care of fixing the rest of languages.

taivlam:
So I was unable to figure out how to add the commit for the Java file, so it's in #57.
(I really wanted to avoid this, but I've haven't more than simple commits in the browser so far.  At a quick glance from web searching, it seems what I originally wanted to do was most easily via the CLI.)

valldrac:
Hmm what a mess. I have never used the browser for pushing commits. I prefer CLI flexibility.

But no worries. I already squashed and merged the pull-requests to master. Thanks for contributing :pray:

PR #61, opened 2021-07-18
Fix depreciated use of ViewModelProviders

benarmstead:
Fixes the warning about using depreciated java classes.


benarmstead:
Fixes the warning about using depreciated java classes.

PR #104, opened 2022-04-26
Add biometric screen lock

### LINKED ISSUE #94: Please add screen lock/unlock option using device fingerprint(Biometrics)

### Feature description

Passphrase encryption is a very useful feature thank you. but during traveling or using mobile in one hand it's not handy to enter the passphrase every time so please add unlock with biometric as an optional additional feature

valldrac:
New feature to lock Molly screen with biometrics for devices with integrated fingerprint sensor and face recognition.

How does it work?

- In privacy settings there is a new option "Screen lock" to enable it.
- The app prompts the user to authenticate with biometrics every time it enters the app. Molly will resume exactly where it was left after passing the validation.
- There is a fixed grace time period of 5 seconds before locking. 
- If any action is started inside the app that sends it to the background (e.g. adding a new contact) the timeout is extended up to 60 seconds.
- Notifications works normally. Only harmless actions via notifications are permitted while Molly has the screen locked: mark as read, or decline a call.
- When passphrase encryption is on, the app requests one biometric pass before checking the passphrase.

Please note:

- It is not intended to replace Molly's passphrase encryption. Biometric screen lock is simply an authentication method to access the app's user interface. There is no data protection or encryption involved. The device owner (or anyone who knows the PIN/pattern/password) could easily override it.
- Although some older devices may have a fingerprint sensor, they may not support the library [androidx.biometric](https://developer.android.com/jetpack/androidx/releases/biometric) that Molly uses.
- The option "Screen security" is required to be active when this feature is enabled.

Closes #94


pixincreate (commented):
@greyson-signal, @cody-signal, could you please look into this and add this feature into the upstream? I believe this is the proper way to add Biometric lock unlike the existing one that has senseless time out thing.  

I'm personally interested in adding this to my fork, but will open up a wide door with  exponentially increasing complexity and never ending conflicts which I'm not interested in as of now.

Issue #128, opened 2022-09-05
Add option to use non-AMOLED dark theme 


BayLee4:
### Is there an existing request for this?

- [X] I have searched the existing issues

### Feature description

Hi,
Molly changed the default background color for the dark theme from [#1B1C1F](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/res/values/material3_colors_dark.xml#L9) to [#000000](https://github.com/mollyim/mollyim-android/blob/main/app/src/main/res/values/material3_colors_dark.xml#L9) (I guess to match the system AMOLED black style, removed since Android 12). Would it be possible to have an option to use the original background color? For some reason I find it more difficult to read with the pure black theme. Thank you!

BayLee4:
Or alternatively and more simply, would reverting the background color to [#1B1C1F](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/res/values/material3_colors_dark.xml#L9) be an acceptable solution?

valldrac:
You're right. Molly should keep the original background color for easy readability. But many Molly users already love the true black theme.

It would be awesome if someone can contribute to this feature. We need to add a 2nd dark theme and develop an script to sync them. A gradle task preferably. This would fix the burden of maintaining an additional theme.

xIUPITERx:
The question is, should we wait until signal implements material you theming?

Cod3dDOT:
Any updates on this?

Uneccessary:
Hey, I personally love the true black background and couldn't live without it, but my friend would like to have original one, how hard would it be to add a 3 option?

ThatOneCalculator:
> The question is, should we wait until signal implements material you theming?

I don't think they'll ever add it... It'd be nice if Molly added Material You, since MD3 is already used

valldrac:
Seems like the AMOLED theme could be improved as well. Someone on [Twitter](https://twitter.com/s8n_ru/status/1764773682428166548) reported it.

Uneccessary:
:eyes:

itsmartashub:
Oh my, what's the reason behind removing the AMOLED theme? Was there too much pressure from users to revert to the original color? Or perhaps you sensed that Signal may soon implement dynamic Material You, so you avoided the hassle and demands from users that could arise from that? Is there any plan to introduce an + OLED theme (again) in the near future, or is that it? Honestly, this is one of the main reasons why I use Molly, deng 😭

itsmartashub:
[...]

PR #148, opened 2022-11-10
Revert dark background color to upstream value for readability purpose

BayLee4:
Linked to issue #128, helps with strained eyes


BHydden:
Adding both options is fine but please don't remove AMOLED true black.

BayLee4:
Having an option only allowing to chose between Dark/AMOLED (no fully fledged themes) shouldn't be a problem to maintain, but I'm no Android dev and don't know how to do it. I'm closing this PR then.

BayLee4:
Fair enough, it just seems that a substantial amount of people actually care about AMOLEDn removing it does not look like the best option then

valldrac:
I agree we need a 3rd theme instead of deprecating th currente true-dark theme. I'm going to close this PR and request for help in the linked issue. Thank you, @BayLee4, for raising this.

PR #157, opened 2022-12-17
Allow registering as a linked device

### LINKED ISSUE #102: Multi Device

### Feature description

A long time ago a signal fork existed which allowed to have a real multi device on Android. 

Could you reintegrate this fork into Molly? The protocol allows it, it would be necessary that when we install Molly we can flash a QR CODE (like on signal desktop) and hop our android is in slave mode.

https://github.com/andrewmattie/Signal-Android-Multidevice

SlugFiller:
Closes #102. Based on provisioning code from signal-cli:
https://github.com/AsamK/signal-cli/blob/master/lib/src/main/java/org/asamk/signal/manager/ProvisioningManagerImpl.java

Synchronizing contacts, groups and profile did not work in testing. But it also did not work when testing with the official Signal Desktop client, so the source of the problem seems to be elsewhere.

Some parts are a bit hacky, because device linking is stateful (requires keeping an open websocket for the entire process), while the normal registration process is stateless (Separate connection can be made for each step).

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
@@ -68,6 +68,7 @@ import org.thoughtcrime.securesms.jobs.PreKeysSyncJob;
 import org.thoughtcrime.securesms.jobs.ProfileUploadJob;
 import org.thoughtcrime.securesms.jobs.RefreshAttributesJob;
 import org.thoughtcrime.securesms.jobs.RefreshSvrCredentialsJob;
+import org.thoughtcrime.securesms.jobs.RefreshOwnProfileJob;
 import org.thoughtcrime.securesms.jobs.RetrieveProfileJob;
 import org.thoughtcrime.securesms.jobs.RetrieveRemoteAnnouncementsJob;
 import org.thoughtcrime.securesms.jobs.StoryOnboardingDownloadJob;
@@ -565,9 +566,14 @@ public class ApplicationContext extends MultiDexApplication implements AppForegr
   }
 
   private void ensureProfileUploaded() {
-    if (SignalStore.account().isRegistered() && !SignalStore.registrationValues().hasUploadedProfile() && !Recipient.self().getProfileName().isEmpty()) {
-      Log.w(TAG, "User has a profile, but has not uploaded one. Uploading now.");
-      ApplicationDependencies.getJobManager().add(new ProfileUploadJob());
+    if (SignalStore.account().isRegistered()) {
+      if (SignalStore.registrationValues().needDownloadProfileOrAvatar()) {
+        Log.w(TAG, "User has linked a device, but has not yet downloaded the profile. Downloading now.");
+        ApplicationDependencies.getJobManager().add(new RefreshOwnProfileJob());
+      } else if (!SignalStore.registrationValues().hasUploadedProfile() && !Recipient.self().getProfileName().isEmpty()) {
+        Log.w(TAG, "User has a profile, but has not uploaded one. Uploading now.");
+        ApplicationDependencies.getJobManager().add(new ProfileUploadJob());
+      }
     }
   }
 
--- a/app/src/main/java/org/thoughtcrime/securesms/DeviceListFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/DeviceListFragment.java
@@ -20,6 +20,8 @@ import androidx.fragment.app.ListFragment;
 import androidx.loader.app.LoaderManager;
 import androidx.loader.content.Loader;
 
+import com.annimon.stream.Stream;
+
 import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
 
@@ -31,6 +33,7 @@ import org.thoughtcrime.securesms.keyvalue.SignalStore;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.thoughtcrime.securesms.util.task.ProgressDialogAsyncTask;
 import org.whispersystems.signalservice.api.SignalServiceAccountManager;
+import org.whispersystems.signalservice.api.push.SignalServiceAddress;
 
 import java.io.IOException;
 import java.util.List;
@@ -45,7 +48,6 @@ public class DeviceListFragment extends ListFragment
 
   private SignalServiceAccountManager accountManager;
   private Locale                      locale;
-  private View                        empty;
   private View                        progressContainer;
   private FloatingActionButton        addDeviceButton;
   private Button.OnClickListener      addDeviceButtonListener;
@@ -66,10 +68,13 @@ public class DeviceListFragment extends ListFragment
   public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bundle) {
     View view = inflater.inflate(R.layout.device_list_fragment, container, false);
 
-    this.empty             = view.findViewById(R.id.empty);
     this.progressContainer = view.findViewById(R.id.progress_container);
     this.addDeviceButton   = view.findViewById(R.id.add_device);
-    this.addDeviceButton.setOnClickListener(this);
+    if (SignalStore.account().isPrimaryDevice()) {
+      this.addDeviceButton.setOnClickListener(this);
+    } else {
+      this.addDeviceButton.setVisibility(View.GONE);
+    }
 
     return view;
   }
@@ -87,7 +92,6 @@ public class DeviceListFragment extends ListFragment
 
   @Override
   public @NonNull Loader<List<Device>> onCreateLoader(int id, Bundle args) {
-    empty.setVisibility(View.GONE);
     progressContainer.setVisibility(View.VISIBLE);
 
     return new DeviceListLoader(getActivity(), accountManager);
@@ -102,16 +106,11 @@ public class DeviceListFragment extends ListFragment
       return;
     }
 
-    setListAdapter(new DeviceListAdapter(getActivity(), R.layout.device_list_item_view, data, locale));
+    setListAdapter(new DeviceListAdapter(getActivity(), R.layout.device_list_item_view, data, locale, SignalStore.account().getDeviceId()));
 
-    if (data.isEmpty()) {
-      empty.setVisibility(View.VISIBLE);
-      TextSecurePreferences.setMultiDevice(getActivity(), false);
-      SignalStore.misc().setHasLinkedDevices(false);
-    } else {
-      SignalStore.misc().setHasLinkedDevices(true);
-      empty.setVisibility(View.GONE);
-    }
+    boolean hasLinkedDevices = Stream.of(data).noneMatch(d -> d.getId() != SignalServiceAddress.DEFAULT_DEVICE_ID);
+    TextSecurePreferences.setMultiDevice(getActivity(), SignalStore.account().isLinkedDevice() || hasLinkedDevices);
+    SignalStore.misc().setHasLinkedDevices(hasLinkedDevices);
   }
 
   @Override
@@ -123,6 +122,10 @@ public class DeviceListFragment extends ListFragment
   public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
     final String deviceName = ((DeviceListItem)view).getDeviceName();
     final long   deviceId   = ((DeviceListItem)view).getDeviceId();
+    if (deviceId == SignalServiceAddress.DEFAULT_DEVICE_ID || deviceId == SignalStore.account().getDeviceId()) {
+      // Sanity check
+      return;
+    }
 
     AlertDialog.Builder builder = new MaterialAlertDialogBuilder(requireActivity());
     builder.setTitle(getString(R.string.DeviceListActivity_unlink_s, deviceName));
@@ -182,11 +185,13 @@ public class DeviceListFragment extends ListFragment
 
     private final int    resource;
     private final Locale locale;
+    private final long   thisDeviceId;
 
-    public DeviceListAdapter(Context context, int resource, List<Device> objects, Locale locale) {
+    public DeviceListAdapter(Context context, int resource, List<Device> objects, Locale locale, long thisDeviceId) {
       super(context, resource, objects);
       this.resource = resource;
       this.locale = locale;
+      this.thisDeviceId = thisDeviceId;
     }
 
     @Override
@@ -195,9 +200,15 @@ public class DeviceListFragment extends ListFragment
         convertView = ((Activity)getContext()).getLayoutInflater().inflate(resource, parent, false);
       }
 
-      ((DeviceListItem)convertView).set(getItem(position), locale);
+      ((DeviceListItem)convertView).set(getItem(position), locale, thisDeviceId);
 
       return convertView;
+    }
+
+    @Override
+    public boolean isEnabled(int position) {
+      long deviceId = getItem(position).getId();
+      return deviceId != SignalServiceAddress.DEFAULT_DEVICE_ID && deviceId != thisDeviceId;
     }
   }
 }
--- a/app/src/main/java/org/thoughtcrime/securesms/DeviceListItem.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/DeviceListItem.java
@@ -5,9 +5,11 @@ import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.widget.LinearLayout;
 import android.widget.TextView;
+import android.view.View;
 
 import org.thoughtcrime.securesms.devicelist.Device;
 import org.thoughtcrime.securesms.util.DateUtils;
+import org.whispersystems.signalservice.api.push.SignalServiceAddress;
 
 import java.util.Locale;
 
@@ -15,6 +17,8 @@ public class DeviceListItem extends LinearLayout {
 
   private long     deviceId;
   private TextView name;
+  private View     primary;
+  private View     thisDevice;
   private TextView created;
   private TextView lastActive;
 
@@ -30,15 +34,19 @@ public class DeviceListItem extends LinearLayout {
   public void onFinishInflate() {
     super.onFinishInflate();
     this.name       = (TextView) findViewById(R.id.name);
+    this.primary    = findViewById(R.id.primary);
+    this.thisDevice = findViewById(R.id.this_device);
     this.created    = (TextView) findViewById(R.id.created);
     this.lastActive = (TextView) findViewById(R.id.active);
   }
 
-  public void set(Device deviceInfo, Locale locale) {
-    if (TextUtils.isEmpty(deviceInfo.getName())) this.name.setText(R.string.DeviceListItem_unnamed_device);
+  public void set(Device deviceInfo, Locale locale, long thisDeviceId) {
+    if (TextUtils.isEmpty(deviceInfo.getName())) this.name.setText(getContext().getString(R.string.DeviceListItem_device_d, deviceInfo.getId()));
     else                                         this.name.setText(deviceInfo.getName());
 
-    this.created.setText(getContext().getString(R.string.DeviceListItem_linked_s,
+    this.deviceId = deviceInfo.getId();
+
+    this.created.setText(getContext().getString(this.deviceId == SignalServiceAddress.DEFAULT_DEVICE_ID ? R.string.DeviceListItem_registered_s : R.string.DeviceListItem_linked_s,
                                                 DateUtils.getDayPrecisionTimeSpanString(getContext(),
                                                                                         locale,
                                                                                         deviceInfo.getCreated())));
@@ -48,7 +56,8 @@ public class DeviceListItem extends LinearLayout {
                                                                                            locale,
                                                                                            deviceInfo.getLastSeen())));
 
-    this.deviceId = deviceInfo.getId();
+    this.primary.setVisibility(this.deviceId == SignalServiceAddress.DEFAULT_DEVICE_ID ? View.VISIBLE : View.GONE);
+    this.thisDevice.setVisibility(this.deviceId == thisDeviceId ? View.VISIBLE : View.GONE);
   }
 
   public long getDeviceId() {
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/account/AccountSettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/account/AccountSettingsFragment.kt
@@ -54,7 +54,7 @@ class AccountSettingsFragment : DSLSettingsFragment(R.string.AccountSettingsFrag
       @Suppress("DEPRECATION")
       clickPref(
         title = DSLSettingsText.from(if (state.hasPin) R.string.preferences_app_protection__change_your_pin else R.string.preferences_app_protection__create_a_pin),
-        isEnabled = state.isDeprecatedOrUnregistered(),
+        isEnabled = state.isDeprecatedOrUnregistered() && SignalStore.account().isPrimaryDevice,
         onClick = {
           if (state.hasPin) {
             startActivityForResult(CreateSvrPinActivity.getIntentForPinChangeFromSettings(requireContext()), CreateSvrPinActivity.REQUEST_NEW_PIN)
@@ -67,8 +67,8 @@ class AccountSettingsFragment : DSLSettingsFragment(R.string.AccountSettingsFrag
       switchPref(
         title = DSLSettingsText.from(R.string.preferences_app_protection__pin_reminders),
         summary = DSLSettingsText.from(R.string.AccountSettingsFragment__youll_be_asked_less_frequently),
-        isChecked = state.hasPin && state.pinRemindersEnabled,
-        isEnabled = state.hasPin && state.isDeprecatedOrUnregistered(),
+        isChecked = state.hasPin && state.pinRemindersEnabled && SignalStore.account().isPrimaryDevice,
+        isEnabled = state.hasPin && state.isDeprecatedOrUnregistered() && SignalStore.account().isPrimaryDevice,
         onClick = {
           setPinRemindersEnabled(!state.pinRemindersEnabled)
         }
@@ -77,8 +77,8 @@ class AccountSettingsFragment : DSLSettingsFragment(R.string.AccountSettingsFrag
       switchPref(
         title = DSLSettingsText.from(R.string.preferences_app_protection__registration_lock),
         summary = DSLSettingsText.from(R.string.AccountSettingsFragment__require_your_signal_pin),
-        isChecked = state.registrationLockEnabled,
-        isEnabled = state.hasPin && state.isDeprecatedOrUnregistered(),
+        isChecked = state.registrationLockEnabled && SignalStore.account().isPrimaryDevice,
+        isEnabled = state.hasPin && state.isDeprecatedOrUnregistered() && SignalStore.account().isPrimaryDevice,
         onClick = {
           setRegistrationLockEnabled(!state.registrationLockEnabled)
         }
@@ -86,11 +86,17 @@ class AccountSettingsFragment : DSLSettingsFragment(R.string.AccountSettingsFrag
 
       clickPref(
         title = DSLSettingsText.from(R.string.preferences__advanced_pin_settings),
-        isEnabled = state.isDeprecatedOrUnregistered(),
+        isEnabled = state.isDeprecatedOrUnregistered() && SignalStore.account().isPrimaryDevice,
         onClick = {
           Navigation.findNavController(requireView()).safeNavigate(R.id.action_accountSettingsFragment_to_advancedPinSettingsActivity)
         }
       )
+
+      if (SignalStore.account().isLinkedDevice) {
+        textPref(
+          summary = DSLSettingsText.from(R.string.preferences__primary_only)
+        )
+      }
 
       dividerPref()
 
@@ -99,7 +105,7 @@ class AccountSettingsFragment : DSLSettingsFragment(R.string.AccountSettingsFrag
       if (SignalStore.account().isRegistered) {
         clickPref(
           title = DSLSettingsText.from(R.string.AccountSettingsFragment__change_phone_number),
-          isEnabled = state.isDeprecatedOrUnregistered(),
+          isEnabled = state.isDeprecatedOrUnregistered() && SignalStore.account().isPrimaryDevice,
           onClick = {
             Navigation.findNavController(requireView()).safeNavigate(R.id.action_accountSettingsFragment_to_changePhoneNumberFragment)
           }
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/ChatsSettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/ChatsSettingsFragment.kt
@@ -7,6 +7,7 @@ import org.thoughtcrime.securesms.components.settings.DSLConfiguration
 import org.thoughtcrime.securesms.components.settings.DSLSettingsFragment
 import org.thoughtcrime.securesms.components.settings.DSLSettingsText
 import org.thoughtcrime.securesms.components.settings.configure
+import org.thoughtcrime.securesms.keyvalue.SignalStore
 import org.thoughtcrime.securesms.util.adapter.mapping.MappingAdapter
 import org.thoughtcrime.securesms.util.navigation.safeNavigate
 
@@ -33,10 +34,17 @@ class ChatsSettingsFragment : DSLSettingsFragment(R.string.preferences_chats__ch
         title = DSLSettingsText.from(R.string.preferences__generate_link_previews),
         summary = DSLSettingsText.from(R.string.preferences__retrieve_link_previews_from_websites_for_messages),
         isChecked = state.generateLinkPreviews,
+        isEnabled = SignalStore.account().isPrimaryDevice,
         onClick = {
           viewModel.setGenerateLinkPreviewsEnabled(!state.generateLinkPreviews)
         }
       )
+
+      if (SignalStore.account().isLinkedDevice) {
+        textPref(
+          summary = DSLSettingsText.from(R.string.preferences__primary_only)
+        )
+      }
 
       switchPref(
         title = DSLSettingsText.from(R.string.preferences__pref_use_address_book_photos),
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/ChatsSettingsViewModel.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/ChatsSettingsViewModel.kt
@@ -35,6 +35,9 @@ class ChatsSettingsViewModel @JvmOverloads constructor(
   }
 
   fun setGenerateLinkPreviewsEnabled(enabled: Boolean) {
+    if (SignalStore.account().isLinkedDevice) {
+      return
+    }
     store.update { it.copy(generateLinkPreviews = enabled) }
     SignalStore.settings().isLinkPreviewsEnabled = enabled
     repository.syncLinkPreviewsState()
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/PrivacySettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/PrivacySettingsFragment.kt
@@ -30,6 +30,7 @@ import org.thoughtcrime.securesms.components.settings.DSLSettingsText
 import org.thoughtcrime.securesms.components.settings.PreferenceModel
 import org.thoughtcrime.securesms.components.settings.PreferenceViewHolder
 import org.thoughtcrime.securesms.components.settings.configure
+import org.thoughtcrime.securesms.keyvalue.SignalStore
 import org.thoughtcrime.securesms.preferences.widgets.PassphraseLockTriggerPreference
 import org.thoughtcrime.securesms.util.CommunicationActions
 import org.thoughtcrime.securesms.util.ConversationUtil
@@ -189,6 +190,7 @@ class PrivacySettingsFragment : DSLSettingsFragment(R.string.preferences__privac
         title = DSLSettingsText.from(R.string.preferences__read_receipts),
         summary = DSLSettingsText.from(R.string.preferences__if_read_receipts_are_disabled_you_wont_be_able_to_see_read_receipts),
         isChecked = state.readReceipts,
+        isEnabled = SignalStore.account().isPrimaryDevice,
         onClick = {
           viewModel.setReadReceiptsEnabled(!state.readReceipts)
         }
@@ -198,10 +200,17 @@ class PrivacySettingsFragment : DSLSettingsFragment(R.string.preferences__privac
         title = DSLSettingsText.from(R.string.preferences__typing_indicators),
         summary = DSLSettingsText.from(R.string.preferences__if_typing_indicators_are_disabled_you_wont_be_able_to_see_typing_indicators),
         isChecked = state.typingIndicators,
+        isEnabled = SignalStore.account().isPrimaryDevice,
         onClick = {
           viewModel.setTypingIndicatorsEnabled(!state.typingIndicators)
         }
       )
+
+      if (SignalStore.account().isLinkedDevice) {
+        textPref(
+          summary = DSLSettingsText.from(R.string.preferences__primary_only)
+        )
+      }
 
       dividerPref()
 
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/PrivacySettingsViewModel.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/PrivacySettingsViewModel.kt
@@ -35,12 +35,18 @@ class PrivacySettingsViewModel(
   }
 
   fun setReadReceiptsEnabled(enabled: Boolean) {
+    if (SignalStore.account().isLinkedDevice) {
+      return
+    }
     sharedPreferences.edit().putBoolean(TextSecurePreferences.READ_RECEIPTS_PREF, enabled).apply()
     repository.syncReadReceiptState()
     refresh()
   }
 
   fun setTypingIndicatorsEnabled(enabled: Boolean) {
+    if (SignalStore.account().isLinkedDevice) {
+      return
+    }
     sharedPreferences.edit().putBoolean(TextSecurePreferences.TYPING_INDICATORS, enabled).apply()
     repository.syncTypingIndicatorsState()
     refresh()
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/advanced/AdvancedPrivacySettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/advanced/AdvancedPrivacySettingsFragment.kt
@@ -183,9 +183,16 @@ class AdvancedPrivacySettingsFragment : DSLSettingsFragment(R.string.preferences
             .append(statusIcon)
         ),
         summary = DSLSettingsText.from(R.string.AdvancedPrivacySettingsFragment__show_an_icon),
-        isChecked = state.showSealedSenderStatusIcon
+        isChecked = state.showSealedSenderStatusIcon,
+        isEnabled = SignalStore.account().isPrimaryDevice
       ) {
         viewModel.setShowStatusIconForSealedSender(!state.showSealedSenderStatusIcon)
+      }
+
+      if (SignalStore.account().isLinkedDevice) {
+        textPref(
+          summary = DSLSettingsText.from(R.string.preferences__primary_only)
+        )
       }
 
       switchPref(
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/advanced/AdvancedPrivacySettingsViewModel.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/privacy/advanced/AdvancedPrivacySettingsViewModel.kt
@@ -63,6 +63,9 @@ class AdvancedPrivacySettingsViewModel(
   }
 
   fun setShowStatusIconForSealedSender(enabled: Boolean) {
+    if (SignalStore.account().isLinkedDevice) {
+      return
+    }
     sharedPreferences.edit().putBoolean(TextSecurePreferences.SHOW_UNIDENTIFIED_DELIVERY_INDICATORS, enabled).apply()
     repository.syncShowSealedSenderIconState()
     refresh()
--- a/app/src/main/java/org/thoughtcrime/securesms/database/loaders/DeviceListLoader.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/loaders/DeviceListLoader.java
@@ -21,7 +21,6 @@ import org.thoughtcrime.securesms.util.AsyncLoader;
 import org.thoughtcrime.securesms.util.Base64;
 import org.whispersystems.signalservice.api.SignalServiceAccountManager;
 import org.whispersystems.signalservice.api.messages.multidevice.DeviceInfo;
-import org.whispersystems.signalservice.api.push.SignalServiceAddress;
 
 import java.io.IOException;
 import java.security.GeneralSecurityException;
@@ -53,7 +52,6 @@ public class DeviceListLoader extends AsyncLoader<List<Device>> {
   public List<Device> loadInBackground() {
     try {
       List<Device> devices = Stream.of(accountManager.getDevices())
-                                   .filter(d -> d.getId() != SignalServiceAddress.DEFAULT_DEVICE_ID)
                                    .map(this::mapToDevice)
                                    .toList();
 
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/AllDataSyncRequestJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/AllDataSyncRequestJob.java
@@ -1,0 +1,74 @@
+package org.thoughtcrime.securesms.jobs;
+
+import androidx.annotation.NonNull;
+
+import org.signal.core.util.logging.Log;
+import org.thoughtcrime.securesms.crypto.UnidentifiedAccessUtil;
+import org.thoughtcrime.securesms.dependencies.ApplicationDependencies;
+import org.thoughtcrime.securesms.jobmanager.Job;
+import org.thoughtcrime.securesms.jobmanager.impl.NetworkConstraint;
+import org.whispersystems.signalservice.api.SignalServiceMessageSender;
+import org.whispersystems.signalservice.api.crypto.UnidentifiedAccessPair;
+import org.whispersystems.signalservice.api.crypto.UntrustedIdentityException;
+import org.whispersystems.signalservice.api.messages.multidevice.RequestMessage;
+import org.whispersystems.signalservice.api.messages.multidevice.SignalServiceSyncMessage;
+import org.whispersystems.signalservice.api.push.exceptions.PushNetworkException;
+import org.whispersystems.signalservice.internal.push.SignalServiceProtos;
+
+import java.io.IOException;
+import java.util.Optional;
+
+public class AllDataSyncRequestJob extends BaseJob {
+
+  public static final String KEY = "AllDataSyncRequestJob";
+
+  private static final String TAG = Log.tag(AllDataSyncRequestJob.class);
+
+  public AllDataSyncRequestJob()
+  {
+    this(new Job.Parameters.Builder()
+                           .setQueue("AllDataSyncRequestJob")
+                           .addConstraint(NetworkConstraint.KEY)
+                           .setMaxAttempts(10)
+                           .build());
+  }
+
+  private AllDataSyncRequestJob(@NonNull Job.Parameters parameters) {
+    super(parameters);
+  }
+
+  @Override
+  public @NonNull byte[] serialize() {
+    return null;
+  }
+
+  @Override
+  public @NonNull String getFactoryKey() {
+    return KEY;
+  }
+
+  @Override
+  public void onRun() throws IOException, UntrustedIdentityException {
+    SignalServiceMessageSender signalServiceMessageSender = ApplicationDependencies.getSignalServiceMessageSender();
+    Optional<UnidentifiedAccessPair> accessForSync = UnidentifiedAccessUtil.getAccessForSync(context);
+    signalServiceMessageSender.sendSyncMessage(SignalServiceSyncMessage.forRequest(RequestMessage.forType(SignalServiceProtos.SyncMessage.Request.Type.CONTACTS)), accessForSync);
+    signalServiceMessageSender.sendSyncMessage(SignalServiceSyncMessage.forRequest(RequestMessage.forType(SignalServiceProtos.SyncMessage.Request.Type.BLOCKED)), accessForSync);
+    signalServiceMessageSender.sendSyncMessage(SignalServiceSyncMessage.forRequest(RequestMessage.forType(SignalServiceProtos.SyncMessage.Request.Type.CONFIGURATION)), accessForSync);
+    signalServiceMessageSender.sendSyncMessage(SignalServiceSyncMessage.forRequest(RequestMessage.forType(SignalServiceProtos.SyncMessage.Request.Type.KEYS)), accessForSync);
+  }
+
+  @Override
+  public boolean onShouldRetry(@NonNull Exception exception) {
+    return exception instanceof PushNetworkException;
+  }
+
+  @Override
+  public void onFailure() {}
+
+  public static final class Factory implements Job.Factory<AllDataSyncRequestJob> {
+    @Override
+    public @NonNull AllDataSyncRequestJob create(@NonNull Parameters parameters, @NonNull byte[] serializedData) {
+      return new AllDataSyncRequestJob(parameters);
+    }
+  }
+}
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/JobManagerFactories.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/JobManagerFactories.java
@@ -90,6 +90,7 @@ public final class JobManagerFactories {
   public static Map<String, Job.Factory> getJobFactories(@NonNull Application application) {
     return new HashMap<String, Job.Factory>() {{
       put(AccountConsistencyWorkerJob.KEY,           new AccountConsistencyWorkerJob.Factory());
+      put(AllDataSyncRequestJob.KEY,                 new AllDataSyncRequestJob.Factory());
       put(AttachmentCopyJob.KEY,                     new AttachmentCopyJob.Factory());
       put(AttachmentDownloadJob.KEY,                 new AttachmentDownloadJob.Factory());
       put(AttachmentUploadJob.KEY,                   new AttachmentUploadJob.Factory());
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/RefreshAttributesJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/RefreshAttributesJob.java
@@ -102,7 +102,7 @@ public class RefreshAttributesJob extends BaseJob {
     String deviceName = SignalStore.account().getDeviceName();
     byte[] encryptedDeviceName = (deviceName == null) ? null : DeviceNameCipher.encryptDeviceName(deviceName.getBytes(StandardCharsets.UTF_8), SignalStore.account().getAciIdentityKey());
 
-    AccountAttributes.Capabilities capabilities = AppCapabilities.getCapabilities(svrValues.hasPin() && !svrValues.hasOptedOut());
+    AccountAttributes.Capabilities capabilities = AppCapabilities.getCapabilities((svrValues.hasPin() && !svrValues.hasOptedOut()) || SignalStore.storageService().hasStorageKeyFromPrimary());
     Log.i(TAG, "Calling setAccountAttributes() reglockV2? " + !TextUtils.isEmpty(registrationLockV2) + ", pin? " + svrValues.hasPin() +
                "\n    Recovery password? " + !TextUtils.isEmpty(recoveryPassword) +
                "\n    Phone number discoverable : " + phoneNumberDiscoverable +
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/RefreshOwnProfileJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/RefreshOwnProfileJob.java
@@ -98,6 +98,8 @@ public class RefreshOwnProfileJob extends BaseJob {
     ProfileAndCredential profileAndCredential = ProfileUtil.retrieveProfileSync(context, self, getRequestType(self), false);
     SignalServiceProfile profile              = profileAndCredential.getProfile();
 
+    SignalStore.registrationValues().clearNeedDownloadProfile();
+
     if (Util.isEmpty(profile.getName()) &&
         Util.isEmpty(profile.getAvatar()) &&
         Util.isEmpty(profile.getAbout()) &&
@@ -112,6 +114,7 @@ public class RefreshOwnProfileJob extends BaseJob {
         Log.w(TAG, "We don't have a name locally, either!");
       }
 
+      SignalStore.registrationValues().clearNeedDownloadProfileAvatar();
       return;
     }
 
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/RetrieveProfileAvatarJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/RetrieveProfileAvatarJob.java
@@ -83,6 +83,7 @@ public class RetrieveProfileAvatarJob extends BaseJob {
 
     if (profileAvatar != null && profileAvatar.equals(recipient.resolve().getProfileAvatar())) {
       Log.w(TAG, "Already retrieved profile avatar: " + profileAvatar);
+      SignalStore.registrationValues().clearNeedDownloadProfileAvatar();
       return;
     }
 
@@ -93,6 +94,7 @@ public class RetrieveProfileAvatarJob extends BaseJob {
         database.setProfileAvatar(recipient.getId(), profileAvatar);
       }
 
+      SignalStore.registrationValues().clearNeedDownloadProfileAvatar();
       return;
     }
 
@@ -122,6 +124,7 @@ public class RetrieveProfileAvatarJob extends BaseJob {
     }
 
     database.setProfileAvatar(recipient.getId(), profileAvatar);
+    SignalStore.registrationValues().clearNeedDownloadProfileAvatar();
   }
 
   @Override
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/StorageSyncJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/StorageSyncJob.java
@@ -170,7 +170,7 @@ public class StorageSyncJob extends BaseJob {
 
   @Override
   protected void onRun() throws IOException, RetryLaterException, UntrustedIdentityException {
-    if (!SignalStore.svr().hasPin() && !SignalStore.svr().hasOptedOut()) {
+    if (!SignalStore.svr().hasPin() && !SignalStore.svr().hasOptedOut() && !SignalStore.storageService().hasStorageKeyFromPrimary()) {
       Log.i(TAG, "Doesn't have a PIN. Skipping.");
       return;
     }
--- a/app/src/main/java/org/thoughtcrime/securesms/keyvalue/AccountValues.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/keyvalue/AccountValues.kt
@@ -329,8 +329,11 @@ internal class AccountValues internal constructor(store: KeyValueStore) : Signal
   val deviceName: String?
     get() = getString(KEY_DEVICE_NAME, null)
 
-  fun setDeviceName(deviceName: String) {
-    putString(KEY_DEVICE_NAME, deviceName)
+  fun setDeviceName(deviceName: String?) {
+    if (deviceName == null)
+      remove(KEY_DEVICE_NAME)
+    else
+      putString(KEY_DEVICE_NAME, deviceName)
   }
 
   var deviceId: Int by integerValue(KEY_DEVICE_ID, SignalServiceAddress.DEFAULT_DEVICE_ID)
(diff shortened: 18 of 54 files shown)


valldrac:
This is very interesting, but it's a poorly documented. I need your help to review it.

Also there are changes in the code, unrelated to device linking, that undo recent commits already in main. What version of Molly are the changes based on?

SlugFiller:
It was originally built on top of dba9f3e53965f89868b42b484c83dc30c8c07c05 then rebased to 1225c56a5ef0933c5bcd96d6ee0967493c38c1e9.

All the changes in the libsignal/service folder are 3-way merged between this repo, the original signal repo, and:
https://github.com/Turasa/libsignal-service-java
I couldn't tell you much about what's in them. You'd have to ask AsamK. The merge was largely automated, with just one pseudo-conflict (changes on different but adjacent lines)

There are a few files that might initially look like they are reverted/unrelated changes. They are not. These are:
/app/src/main/java/org/thoughtcrime/securesms/jobmanager/migrations/PushProcessMessageQueueJobMigration.java
/app/src/main/java/org/thoughtcrime/securesms/messages/MessageContentProcessor.java
/app/src/main/java/org/thoughtcrime/securesms/messages/MessageDecryptionUtil.java
/app/src/main/java/org/thoughtcrime/securesms/util/GroupUtil.java

The changes in these files is to adjust for an API change in libsignal/service which allows group V1 and group V2 to co-exist in the API. The class `SignalServiceGroupContext` is added as a form of "union class" of `SignalServiceGroup` (v1) and `SignalServiceGroupV2`. Various methods that would previously return `SignalServiceGroupV2` are changed to return `SignalServiceGroupContext` instead. I'm not sure if v1 groups are necessary for provisioning, but it's easier to add support for the dual-version API than to cherry pick the changes from the library.

There is another change in the API, that I ended up applying in the opposite direction (i.e. reverting the API change). That is changing the definitions of device id parameters from being sometimes `int` and sometimes `long` to always be `int`. Since the max for device id is 4 anyway, this change makes sense. But applying it application-wide would require changing code related to message decoding, or adding a heck of a lot of superficial long-to-int typecast.

Once you set aside these changes, what's left (i.e. The parts I actually wrote) should be far easier to understand. I've made comments on the significantly hacky parts. If you have any specific questions about any of the other parts, I'll do my best to answer.

valldrac:
> All the changes in the libsignal/service folder are 3-way merged between this repo, the original signal repo, and:
> https://github.com/Turasa/libsignal-service-java
> I couldn't tell you much about what's in them. You'd have to ask AsamK. The merge was largely automated, with just one pseudo-conflict (changes on different but adjacent lines)

Turasa/libsignal-service-java has dozens of commits and bugfixes that are not related to this pull-request. It is not wise to automatically merge them into Molly. This is a critical component under heavy development by Signal. Also, the latest commit is from Nov 7th and Molly's libsignal is newer. There's no way I can review this and maintain it later.

You would need a clean, manually curated merge, that includes only the minimal set of changes required to register as a linked device.

SlugFiller:
Okay, I made a best effort to reduce the changes to only what is absolutely necessary for device linking. Linking and, at the very least, sending a note to self, is tested and still working.

It is now far fewer changes, and should be easier to read.

SlugFiller:
Fixed lint errors

valldrac:
Thank you. Now it's much neater. There are still some leftovers, but it's not a problem.

I'd like to try this out right away before discussing the implementation details. We have a dedicated testing group and they will love to try this out. I can build one testing APK with your code merged. Would you consider joining our [testing channel](https://matrix.to/#/#mollyim-testing:matrix.org) in Matrix? I'm in that channel too.

SlugFiller:
I don't have a Matrix account, and am even less likely to get one after they removed guest access support, while showing the same sort of attitude that convinced me to make this pull request here and not upstream to Signal. I would sooner create a Discord account.

I am open to any sort of WebIRC, though.

Oh, and good luck to all the testers. I hope they have more luck synchronizing contacts than I did.

valldrac:
APK ready for testing: https://github.com/mollyim/mollyim-insider-android/releases/tag/v6.3.6-1%2Blink-device

SlugFiller:
Added possible fix for contact and group sync (requires testing). Also, made it so the pin setting popup doesn't appear, since it's impossible to set pin from the linked device (It instead gets the master key directly from the primary).

SlugFiller:
Trying to interact with the testers on the testing channel, but it seems like they're not seeing my messages. Must be a Matrix bug.

So I'll just point out here that one thing in particular that I want tested, is sync of blocked contacts. This should absolutely work, even with the previous version. And if it's not working, then it means that there's a serious issue that needs to be investigated.

In particular, I want the following process tested:
1. Ready primary account
2. Block some contacts on primary
3. Link secondary
4. Check blocklist on secondary. It should match primary.

I only have a limited ability to test this alone (I don't have a bunch of Android phones lying around), so input from testers is valuable.

valldrac:
> Added possible fix for contact and group sync (requires testing).

Great. I'm building a new testing APK.

> Trying to interact with the testers on the testing channel, but it seems like they're not seeing my messages. Must be a Matrix bug.

I haven't received any message from you since you joined the room. May rejoining fix it? Also Element Web seems to work fine most of the time. 

> I only have a limited ability to test this alone (I don't have a bunch of Android phones lying around), so input from testers is valuable.

You can try emulators. That's what I use for testing. If you need extra phone numbers for registering in production servers, please let me know.

Up to now, the testers have identified the following issues and improvements:

* [ ] In the link device screen, allow tap to copy the QR link
* [ ] Add an indicator of main/secondary device in Settings' link-device list
* [ ] Disallow account deletion action in Settings on linked device
* [ ] Linking from secondary devices, gives a network error, the same happens if trying to link a second secondary from the primary device
* [ ] An extra sync (empty) message has been inserted in conversation

valldrac:
Debug log for the extra message issue:

https://debuglogs.app/android/6.3.6/5f0a809d47011792659b548e5994cd9a8c4414c7fa9928551411cdd880c2007b

Message-ID: 1671545932600

image

image

SlugFiller:
The Matrix issue probably stems from the fact that I'm not using a matrix.org user. There's probably a bug in federation. That's my best guess, anyway.

The empty sync job is a bit difficult to debug. First, because it appears random. Second, because the log contains multiple attempts, calls, even a re-register as primary. Without at least a timestamp narrowing it down, I can't even tell where in the log I need to look.

The log does give some interesting messages that can be investigated regardless, though. I can see the part where it fails reading group and PNI sync messages. The prior is a bit unexpected, because those should only transmit for V1 groups. I'm also seeing indication that contact and key sync messages are being read okay. That's a good sign, and I'd like to know if, combined with the above fix to storage, it allows a full sync.

Registering a second secondary from primary should, by all means, work. My own device is a third secondary, since I have two desktop clients attached. My guess here is that perhaps the tester did multiple link attempts, but did not delete previous linked devices from the primary. This could have caused it to reach the maximum of 3 linked devices. If that's not the case, I'm going to need the error timestamp in the log.

I'll look into the others since they are simple UI changes.

I assume the reason to remove account deletion from a linked device is because it doesn't work, and can only be done from primary. Besides account deletion, the effects of device deletion need to be investigated as well. As in, in the device list if you tap on a device, it offers the option to delete it. If this doesn't work from a linked device, the UI option should be removed. If it does work, then the effects of doing a self-delete need to be investigated, in case it outright crashes the client or something.

clauz9:
> Without at least a timestamp narrowing it down, I can't even tell where in the log I need to look.

`Message-ID: 1671545932600`  isn't this a timestamp ?  

do you need the message after it ? or before it ?

SlugFiller:
@clauz9 Ah yes. This helps. Nothing immediately stands out, but I'll compare it with the other messages, and see if there's any noticeable difference.

While looking at the account settings, I noticed another thing that needs checking: Device transfer. The device transfer process copies messages and contacts, but not the registration. Instead, the receiving device is sent through a re-registration process.

So, things to check: Is it possible to link a device as part of the re-registration process? Does the link complete safely and correctly? Is "transforming" a linked device to a primary via transfer a particularly bad idea? More plainly, should the "device transfer" option be removed from linked devices?

If it should be, then the entire "Account" section of the settings can be removed, since pin and phone changes can't be done from a secondary

valldrac:
> While looking at the account settings, I noticed another thing that needs checking: Device transfer. The device transfer process copies messages and contacts, but not the registration. Instead, the receiving device is sent through a re-registration process.

You're right. Note that device transfer is based on the backup & restore code. Basically, device transfer streams the backup file from one device to another, using Wi-Fi direct, and then the backup is restored in the target device. Like backups, it will require re-registration, as account credentials are not saved.

> If it should be, then the entire "Account" section of the settings can be removed, since pin and phone changes can't be done from a secondary

If this is the case, we can replace the Account section by a help text indicating the account operations should be performed in the primary device instead.

SlugFiller:
Added copying the QR link by tapping on the text. Could have probably done the same for tapping on the QR, but seems redundant.

Fixed an issue where the device name would be reset to null after a while, due to a capabilities update re-sending a blank name.

Redesigned linked devices page. Now shows the primary device with the tag "Primary", and tags the current device with "This device". Tapping on either does nothing. Unlinking a device from a linked device needs to be tested. If it doesn't work, I'll remove the option completely.

Also needs testing is whether messing with Chats->Generate link previews, and Privacy->Mesaging has any effect, since in desktop, they cannot be modified. If modifying them does not work, they should be disabled.

I've noticed an issue when linking where it resets the avatar. The is probably caused by the fact that RefreshOwnProfileJob starts a RetrieveProfileAvatarJob asynchronously, which causes a race condition with ProfileUploadJob, that is called by basically everything every other second for some reason. I need to figure out a work around to completely lock out profile uploads until downloading the profile is confirmed complete.

Still haven't heard any test results for block-list sync, or fixed contact/group sync.

SlugFiller:
Fixed profile and avatar download on link. Now guaranteed to happen before a profile upload.

valldrac:
Thank you! I released another Insider with the changes.

> Still haven't heard any test results for block-list sync, or fixed contact/group sync.

It was reported in Matrix. I think the fixes passes all the tests.

There are more feedback in the testing room as well. We should decide whether to move all the conversation here, and tell the testers so, or try to fix the issue with Matrix. I would say try the later first but let me know what you think.

xyz-nobody:
Thanks I tested all the features with signal-cli in master or another android in master, and it works perfectly.


I just noticed (and this is normal) that when we use parameters for the profile that are not present in the master, then the profile of the slave is always deleted.

Ex : Privacy 
Who can ...
See my phone number : Nobody
Find me by phone number : Nobody 

But I think it's normal because my signal-cli doesn't support these settings yet.

While when the Android version supports it in master, the slave is well synchronized, except for the parameter : Find me by phone number : Nobody 

But ditto, this must be normal because it is not yet a parameter officially supported by Signal.

In any case a huge thank you, it's a really important feature that you have enabled. I still don't understand why Signal never implemented it on Android.

Anyway, thanks guys

SlugFiller:
Okay, this is a big one: Added support for changing phone number from primary

WARNING: HIGHLY EXPERIMENTAL. The code is 90% copy pasta, and 10% absolute guesswork.

When a primary changes a phone number, it regenerates the PNI and all related encryption keys. These are necessary for inter-device communication, so the number change request must be accompanied with encrypted messages containing the new keys, to be sent to each device. If the list of messages does not match the list of linked devices stored on the server, the request fails with mismatched devices error, and must be re-crafted.

Any linked device must process the message and update its internal encryption keys to match the sent ones. Both signal-cli and Signal Desktop do it in a largely similar fashion, but Molly/Signal Android stores the keys in a very different way, so adjustments needed to be made accordingly.

@valldrac "Fixing the issue with Matrix" would, at best, require a discussion with the matrix.org admins and/or the admins of my homeserver, and at worst, a pull request. And, for obvious reasons, I'm not planning on juggling two codebases at once.

I already need the bottom line, really. Like, if you tell me "Setting pin and transferring account does not work from linked device, change the account page to a text saying 'abcdef'", and maybe put a checkbox next to it, then I'll just do it.

If most of the fixes are already working, then all I have left is to disable/grey out a few primary-only features in settings, and give one more shot at pin-less group sync, for signal-cli users, since signal-cli doesn't support pin storage sync yet.

@xyz-nobody Those options don't appear for me, probably because I'm using the free version. But "See my phone number" is shared through pin storage sync's account record, so signal-cli is not going to be able to interact with it, for the simple reason that it doesn't have pin storage support. On the other hand, it does mean that the setting is two-way for Android primary.

By contrast, "Find me by phone number" is NOT in the pin storage. It's actually set in the account attributes call, which contains various device capabilities, including whether stories or sender key are enabled. It's not clear if a linked device can even set those, or if they're supposed to be common with the primary, or different per-device (device name is one of them, which suggests they may be per-device).

This is separate from configuration sync which contains read receipts, typing indicator, unidentified delivery indicator, and link previews. Those are synchronized primary to linked only, using a simple sync message.

It's not clear why Signal decided to have 3 different ways to sync configurations. Even if some are newer, it would have made sense to simply port everything to latest.

P.S. I can explain why Signal never implemented this on Android (and also why this pull request is here and not upstream), but it's basically internet drama and ego wars. While some fine-tuning was done to fit the latest version, much of the code in this patch is copied or adapted from patches created back in 2016. If it was a priority, anyone could have gotten at least the basic QR-scan link process working. Even with all the adjustments made so far, I've been working on this thing around a month and a bit. That's all it takes.

SlugFiller:
Disabled v1 configuration items, and added appropriate flavor text indicating they can't be modified from a linked device.

SlugFiller:
Removed request for v1 groups. Looking at `MultiDeviceGroupUpdateJob`, it has a very clear `if (record.isV1Group()) {`. Since support for v1 groups has been removed from the protocol library, and since v1 groups will almost always be auto-migrated to v2, there's no point in attempting it. v2 groups can only be synchronized via pin storage.

SlugFiller:
Rebased to latest

SlugFiller:
I've recently tested re-linking after unlinking from main. And while the popup text might use some flavoring for linked device (it currently says you might have registered from a different device) it works fine.

So, far as I'm concerned, this PR is 100% waiting on test results. Currently needs testing:

- [ ] Does anything in Settings->Account work from a linked device?
- - [ ] Create a PIN and other PIN settings
- - [ ] Change phone number (theoretically should not work)
- - [ ] Transfer account
- - [ ] Delete account
- [ ] Does changing numbers from main work? Does the linked device continue to work, or gets unlinked?
- [ ] What happens if you link two devices, and try to unlink one of the devices from a linked molly? Does the unlink succeed?

After I have results for the above and make appropriate adjustments, this PR should be merge ready

xIUPITERx:
Can we help with the testing?

SlugFiller:
@xIUPITERx There's a test build [here](https://github.com/mollyim/mollyim-insider-android/releases) and a Matrix room for discussing tests at #mollyim-testing:matrix.org (Although I am unable to post in the latter, because Matrix is buggy).

xIUPITERx:
1. I think the PIN settings should also be locked on the linked device. Currently I see conflicting PIN settings (signal on main device vs. linked Molly) for the same account.
image
2. The main device is listed under unnamed device, because you can only set a name on the linked devices, could something be done about that?
image

clauz9:
I got a crash after I entered the  SMS reg code, on a Lineage OS 19.1 

log: https://debuglogs.app/android/6.6.3/10157403f97dd0dc1bc2b9c32b9ee0caecb998ca1eff770c061eb64fb354ca9f

SlugFiller:
Fixed the crash bug when registering as primary. Should have probably mentioned that testing functionality as a primary is also necessary.

@xIUPITERx There's no part of the protocol where the primary sends the pin to a linked device, but it does send a pin-equivalent key. It's possible for the linked device to detect that it has this pin-equivalent key, but it's not clear if simply knowing that it exists is enough to manage the settings. More importantly, it's not clear whether the linked device can manipulate these settings at all.

So, one step would be to try to skip pin creation on the primary, and try to create it on the linked device instead. Another option is to use the advanced pin options to try to disable the pin.

I could also create a build that forces "pin reminders" and "registration lock" to be enabled, at least if the linked device detects having a key (It's a one line change in `app/src/main/java/org/thoughtcrime/securesms/components/settings/app/account/AccountSettingsViewModel.kt`). But it's not clear whether they work from a linked device in the first place.

Regarding the primary device name, it's a problem. If the primary device line is not shown, there's no clear way to indicate that the primary device is the current device (It's possible to imply it via no other device being current, but if the device list gets long, this could be hard to verify). It might be possible to add a way to set it for the primary device. I believe there's a protocol for changing a device's name after it's been created, but I'm not sure.

It's also possible to change "Unnamed device" with something that's less of an eyesore. Maybe remove the device name line completely, or force the name to "Primary" only for unnamed primary devices. Note that it's not strictly impossible for a linked device to be unnamed (I've created a few such during testing). While the current linked client asks for a name in advance, it's entirely possible for alternative clients to skip that phase and leave the device unnamed.

Anyway, it's a design call. There's not just a single solution. And, the current situation is also one of the options.

xyz-nobody:
According to my tests a slave can do everything.
Including touching the PHONE_NUMBER_PRIVACY parameters:

Who can see my phone number : 
Find me by phone number : 

On the other hand, as soon as the master connects, it will overwrite the changes that the slave has made.
So as long as the master is not connected, the slave has the hand on everything, but as soon as the master arrives, the modifications of the slave are overwritten

I tried to make scenarios with several, slave, or I disconnect etc., all goes well.

Just that according to my test it is the master that has the last word, so I think it must be the same for the pin

valldrac:
> It's also possible to change "Unnamed device" with something that's less of an eyesore.

I suggest to display a name based on the device ID for "unnamed" devices always, either they are primary and secondaries. For example, using the ordinals ("1st device") or the ID in the name ("Device #0").

> Fixed the crash bug when registering as primary.

New test build with the fix [here](https://github.com/mollyim/mollyim-insider-android/releases/tag/v6.6.3-1%2Blinkdevice.rev5).

SlugFiller:
Rebased. Made unnamed devices less of an eyesore. Made account pin settings try to sync with primary on whether a pin is set or not. Can revert if none of the pin-related stuff work anyway.

clauz9:
The crash is still there, in rev5 

https://debuglogs.org/android/6.6.3/7ce7081cd8a673f5b6d011852daf70d781ec7a2e0ced3c7dfd5e0dd03c0dd918

valldrac:
> Can revert if none of the pin-related stuff work anyway.

At this point it can be useful to add new changes as independent commits instead of amending the current one. I can squash them all at merge time.

SlugFiller:
Crash fix, attempt number 2.

>At this point it can be useful to add new changes as independent commits instead of amending the current one. I can squash them all at merge time.

It's easier for me to amend. Makes rebasing easier too. In the rare case that I absolutely need to check a previous attempt, I have the reflog references for that.

And also, while not shown as a log, all the commits are technically recorded in this thread.

valldrac:
> Crash fix, attempt number 2.

Build to test it [here](https://github.com/mollyim/mollyim-insider-android/releases/tag/v6.6.3-1%2Blinkdevice.rev6).

SlugFiller:
That was quite a gotcha from lint for adding a parameter to a translated string. Changed the name of the string so it doesn't clash with the translations (Should probably also remove the string from translations, but there's like a thousand unused strings in translations currently, so it would be better to eventually do a single big bookkeeping on all of them).

clauz9:
Looks fixed... 👍

SlugFiller:
Nevermind, lint screamed at me again. Let's see if I got it this time.

xIUPITERx:
I can not add a new device, previously I tried it with WSA on Windows (there it worked), but now on the smartphone it no longer works.
The error message "Network error" always appears.

clauz9:
When I'm linking a device, I probably don't need the new account onboarding cards ....

clauz9:
I got a crash, I was linking an Android, I scanned the QR code, and then tapped, back on the device I was linking... 

then it crashed...
```
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException org.whispersystems.signalservice.api.push.exceptions.PushNetworkException: java.io.InterruptedIOException: interrupted
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.getServiceConnection(PushServiceSocket.java:1797)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1708)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1648)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1642)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.finishNewDeviceRegistration(PushServiceSocket.java:434)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.api.SignalServiceAccountManager.finishNewDeviceRegistration(SignalServiceAccountManager.java:845)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.registration.LinkDeviceRepository.attemptDeviceLink$lambda$2(LinkDeviceRepository.kt:59)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.registration.LinkDeviceRepository.$r8$lambda$E4Vpy1o8X7QPBkuPUpv9540LZHY(Unknown Source:0)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.registration.LinkDeviceRepository$$ExternalSyntheticLambda0.call(Unknown Source:8)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at io.reactivex.rxjava3.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at io.reactivex.rxjava3.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:38)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at io.reactivex.rxjava3.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:25)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at java.lang.Thread.run(Thread.java:919)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException Caused by: java.io.InterruptedIOException: interrupted
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okio.Timeout.throwIfReached(Timeout.kt:116)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okio.OutputStreamSink.write(Okio.kt:62)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okio.AsyncTimeout$sink$1.write(AsyncTimeout.kt:106)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okio.RealBufferedSink.flush(RealBufferedSink.kt:222)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http1.Http1Codec.finishRequest(Http1Codec.java:166)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:84)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:127)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.net.DeviceTransferBlockingInterceptor.intercept(DeviceTransferBlockingInterceptor.java:38)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.net.DeprecatedClientPreventionInterceptor.intercept(DeprecatedClientPreventionInterceptor.java:36)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.net.RemoteDeprecationDetectorInterceptor.intercept(RemoteDeprecationDetectorInterceptor.java:22)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.thoughtcrime.securesms.net.UserAgentInterceptor.intercept(UserAgentInterceptor.java:20)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:257)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at okhttp3.RealCall.execute(RealCall.java:93)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	at org.whispersystems.signalservice.internal.push.PushServiceSocket.getServiceConnection(PushServiceSocket.java:1790)
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 	... 17 more
[v6.6.3-1+linkdevice.rev5-FOSS] [54187] 2023-01-14 15:57:17.259 GMT+02:00 E SignalUncaughtException 
```

SlugFiller:
@xIUPITERx Without more info, my primary guess would be that you've linked too many devices, and didn't delete/unlink previous devices. There's a limit on the maximum number of linked device, and it's rather low. So it's entirely possible that you'd link a desktop, and immediately run out of the limit to register an Android.

@clauz9 By "new account onboarding", do you mean the stuff at the bottom? At the very least the "create PIN" megaphone is force-disabled for linked devices. While the other ones are probably not needed, it's entirely possible that they might be useful, even on a linked device. For example "Turn on stories" is absolutely something you can do on a linked device, even if most sane people wouldn't want to.

The linked device should detect when a PIN exists, since it must receive the master key for storage sync. If the linked device fails to see a PIN set on master, it would also fail to sync groups. In what particular way/screen did the linked device show PIN unset? Are we talking about the Account settings?

It does sound like PIN settings don't really do much from the linked device, though, especially if enabling/disabling reg lock has no effect.

I pushed a possible fix to the crash bug when hitting back during a successful link. But because this changes the error handling, it's also necessary to check out time out for scanning QR (occurs if you do not scan the QR for a full minute after entering the device name).

xIUPITERx:
> @xIUPITERx Without more info, my primary guess would be that you've linked too many devices, and didn't delete/unlink previous devices.

It was A new Account without any linked devices.
What information do you need?

clauz9:
>I should automate the release of insider builds for feature branches like this. Now the releases requires some manual steps on my part.

@valldrac  did you do this ? 

> By "new account onboarding", do you mean the stuff at the bottom?

yes 

> Are we talking about the Account settings?

yes, the account settings menu doesn't update, to reflect the changes 

> it's entirely possible that they might be useful, even on a linked device. For example "Turn on stories" is absolutely something you can do on a linked device, even if most sane people wouldn't want to.

I would turn stories on, I don't mind them.

I was saying you don't need those cards because since you're linking a device, this means you've already seen them on your primary device, and presumably you already know what you're doing.... 

but I suppose one could create custom cards, specific to linked devices 🤔 to outline what one ca do with linked devices that is special.. 

>At this point it can be useful to add new changes as independent commits instead of amending the current one. I can squash them all at merge time.

I would take this advice from @valldrac,  it would make it clearer what has been addressed and what still needs to be addressed. 

@SlugFiller

SlugFiller:
@clauz9 Hmm, showing PIN presence on the account settings is finicky for me too. Sometimes it shows as not set, but if I exit the screen, the go back into account settings, it shows as set. I should probably investigate this, but if the settings make no difference, then there's no point.

> I was saying you don't need those cards because since you're linking a device, this means you've already seen them on your primary device, and presumably you already know what you're doing....

Not necessarily. I haven't investigated it deeply, but I think the "turn on stories" one only appears if you actually have stories turned off. It's also possible that your main is not Android, and therefore, does not have the same cards. It could be signal-cli. And, when looking at Signal-Desktop's code, I saw what looks like a code path for registering it as a main device.

Now, if one of those cards does not actually work, and/or crashes the client if you try to use it, THAT would be a reason to remove it.

SlugFiller:
Tried to rebase to latest, but failed to compile due to an issue introduced in 3a1f978ef8278569c75ae5cfc81301d02e98e841

```
/molly/app/src/main/java/org/thoughtcrime/securesms/backup/FullBackupImporter.java:278: error: incompatible types: possible lossy conversion from long to int
       preferences.edit().putInt(preference.getKey(), preference.getIntegerValue()).commit();
```

This is due to the fact that `integerValue` is defined as uint64 in the protocol, but `SharedPreferences.Editor.putInt` expects an integer. I'm kind of wondering how validation didn't pick it up sooner

valldrac:
> Tried to rebase to latest, but failed to compile due to an issue introduced in [3a1f978](https://github.com/mollyim/mollyim-android/commit/3a1f978ef8278569c75ae5cfc81301d02e98e841)

I fixed it in the `molly-6.8` branch. You can rebase to that branch meanwhile.

The problem is that I've changed the process of merging upstream code. Now I do it in a dedicated branch. Then when it's well tested, I merge it into `main` to release a stable version. But this bug was left in `main`.

I have to refrain myself from committing directly to `main`.

SlugFiller:
@valldrac Hmm. I'll wait for the merge. I don't want to base on a branch. I'm not getting very meaningful test results anyway, so it's unlikely that this PR will be ready to merge soon.

SlugFiller:
Rebased and squashed. Not that there was much to. Haven't tested, though, so... Use with care.

clauz9:
> Rebased and squashed. Not that there was much to. Haven't tested, though, so... Use with care.

Why did you squash it ? 🤔  you can leave the commits visible so we can track the progress, let valldrac squash them when he merges into the main branch...

SlugFiller:
@clauz9 If I wanted a perfectly honest history, I'd merge instead of rebasing. The rebase changes are more numerous than the ones squashed anyway.

I prefer squashing myself, and keeping the PR a fast-forward, and avoid "Authored by Commited by" mess in the main branch after merge.

The contents of this PR thread are more reliable for tracking progress, which, incidentally, hasn't changed significantly since I posted [this](#issuecomment-1371215644).

valldrac:
> Rebased and squashed.

New build: https://github.com/mollyim/mollyim-insider-android/releases

> The contents of this PR thread are more reliable for tracking progress, which, incidentally, hasn't changed significantly since I posted [this](#issuecomment-1371215644).

Sorry if it doesn't move forward. I don't have a lot of time to coordinate the tests, so I refer everyone interested in this feature to talk to you directly in this thread.

What are your expectations about this PR getting accepted and merged?

SlugFiller:
@valldrac

>What are your expectations about this PR getting accepted and merged?

Well, the list is pretty self-descriptive, but I can elaborate on it as two main categories:

1. Things that work in primary but might not work in a linked device. It doesn't do any harm for those to stay there, but if these are not caught by the testers, they will eventually be caught by users, and become a new issue report ("I tried to X from a linked device and it didn't work"). My expectation for each of those would be "I did XYZ on linked device, then tried to W from primary/same device/other linked device, and all that happened was V".

2. The change phone number feature. This MUST be tested in staging, because of the potential to harm the user if they need to use it, and it turns out to not work. I would say this one is the higher priority, as the other category would merely be occasionally annoying for a few users, but largely on the level of a documentation issue.

I hope this answers your question, assuming I understood it correctly.

I suppose I could list individual testing steps for each of the list items, if they are not self-explanatory.

pixincreate:
Will this allow device sync-data to be backed-up when chat backup is done?

xIUPITERx:
I am not sure if that happened because of molly.
The main signal app had problems with that a few weeks ago.

valldrac:
For all those interested in testing this feature, just remember that we have experimental pre-builds of Molly ready for download. Look for "linkdevice" releases at https://github.com/mollyim/mollyim-insider-android/releases.

johanw666:
I don't kno if this qualifies as a bug: I linked Molly linkdevice rev6 to a Signal on the same phone. I screenshotted the qr code, uploaded it to my laptop and switched to Signal to read it. Tht worked. When I switched back to Molly it hung on the QR code screen, it somehow had not received the notification that linking had succeeded. when I closed Molly and opened it again it asked for a PIN, when I didn't supply one it then worked OK as linked device, when I sent a message frm Signal it received the message.

Debuglog: 




Image attached to message
Image attached to message
Image attached to message
Image attached to message
https://debuglogs.org/6abd0cd36c2493b176cd3e7827b1f9f4bdca4c56d522bb5e56276d97dbaf3328
Image attached to message
The recording of Fysica 2020 can be found at https://www.youtube.com/watch?v=SDnlUQPxup4&feature=youtu.be
https://debuglogs.org/c23ae9b453c4a51fde277cfde71fd0c0801daa02e0e3bf523e7aa1a961588efc

Test
https://debuglogs.org/f4d87cb41a8167edc7e5fe8af7f651d6c1f984458b601774b5195a8531a6849e

https://www.youtube.com/watch?v=Id-M1lerGDQ
https://debuglogs.org/4f8711033075ec0a5bc2237023ba7c173d447375c00f1c641c500af9a60b7a34
https://debuglogs.org/44413fa237156d45641f184ede40ec3069341ef66cf330a5ac493a395b0a77ea
Crash: https://debuglogs.org/213354c10f4c2e65faa550e2e11621e26a820d71a0e3e0e9882f48d81b7ac655
Test
Tez
https://debuglogs.org/android/6.9.2-1.0-JW-linkdevice/1fcf33b0a409e1f39fb4ff717b1c5bce27e1042b5369a5256f5dabcfe700b0a4

SlugFiller:
@pixincreate Re-installing (and, in the process, re-registering) the main device unlinks all the edge devices, because during linking they actually do a key exchange to be able to communicate with each other securely, and the master device's keys are regenerated during re-registration. This is also why the "secret number" for all contacts changes, causing them to become unverified. The difference is that with contacts, you can simply re-verify them after the re-registration. With linked device, the link process IS the verification. You don't want your messages automatically shared with an unverified device. That's why linking has to be re-done.

@johanw666 I'm not too familiar with pinned conversations. I assume they are synchronized through pin storage, as that is the only way for a linked device to send things to the main. "People who are no longer registered" is a weird use case. But at any rate, if the problem exists on desktop, it's pretty likely to be present on linked Molly as well, since that's just a protocol flaw, not something that is implementation specific.

The linking process does not really support switching to another app, for the simple reason that the linking process requires a websocket to be open through the entire process, and it's completely unpredictable what Android would do with the websocket if you switch away (It is allowed to completely kill the app, after all). For the same reason, linking also necessarily times out after 1 minute, since the server force-closes the socket after that. "Freezing" seems to be odd, since the expectation would be that the app would be killed and recreated, causing a new QR to be generated. It's pretty hard to predict how an app would behave in the background.

Asking for a pin is a bit weirder, since there's a section that explicitly prevents that if the device has a device id other than 1. Since setting the device id is one of the first steps of the registration, it shouldn't have arrived at that condition, unless you're using a very old version of the patch, or something really weird happened with the data storage due to the app being killed unexpectedly.

johanw666:
> @johanw666 I'm not too familiar with pinned conversations. I assume they are synchronized through pin storage, as that is the only way for a linked device to send things to the main. "People who are no longer registered" is a weird use case. But at any rate, if the problem exists on desktop, it's pretty likely to be present on linked Molly as well, since that's just a protocol flaw, not something that is implementation specific.

The problem exists less on desktop. When I pin those conversations on desktop they automatically unpin, and that is synced to Android. When I only pn them on  Android, desktop ignores them.

The use case is that it is a conversation with someone who used to have a Signal account but is passed away and I don't have access to the number or acount, but I do like it to remain pinned in Signal Android.

pixincreate:
> @pixincreate Re-installing (and, in the process, re-registering) the main device unlinks all the edge devices, because during linking they actually do a key exchange to be able to communicate with each other securely, and the master device's keys are regenerated during re-registration. This is also why the "secret number" for all contacts changes, causing them to become unverified. The difference is that with contacts, you can simply re-verify them after the re-registration. With linked device, the link process IS the verification. You don't want your messages automatically shared with an unverified device. That's why linking has to be re-done.

So there's no other way work around for this unless the basic logic of this is modified/changed. Or if we can add a reference key to the backup so that it remembers the sync device on restoring the backup to keep the devices in sync, would be useful, right? By this, it would open up another possibility to restore chats between devices if I'm not wrong here.

johanw666:
> Should Molly only be used as a linked device if the main device also has Molly installed, or is a normal Signal installation on the main device also OK?

It works if the other device has Signal, I managed to make it work even if Signal and Molly-linkeddevice are on the same phone. I screenshotted the QR code, uploaded it to a laptop and read that code with Signal. In the later case Molly seemed to hang on the QR code when I returned it from the background but after a restart it worked as a linked device.

SlugFiller:
@pixincreate Storing a list of linked devices in the backup wouldn't help. They're already stored on the server. There's simply no secure way to communicate with them, since neither the linked device, nor the primary, have a way of telling if any message came from a real device, or is spoofed. Storing any sort of keys in the backup would be a security risk, since a backup is only secured by a user-generated passphrase (which can be blank), and leaking a key is 100 times worse than leaking your entire chat history.

During the linking process, the linked device generates a one-time encryption key, that is NOT sent to the server, and is instead encoded into the QR. The main device scanning the QR uses this key to send its own keys, via the server, to the linked device. Since the server never receives the linked device's key, it is also unable to decode the primary device's key. Therefore, the server is left with no keys, and the link is secure. There is no way to automate this process, because there's no way to pass the one-time key securely, except via QR. That's why it has to be repeated every time the keys are lost.

The one exception to this is that during phone number change, the keys are changed, but the old keys are not lost, so the primary device uses the old keys to transfer the new keys. This doesn't work during backup, because the primary device can't predict what the new keys would be during backup, nor can it keep the old keys until restore.

@johanw666 It seems to be a feature, not a bug:
```java
  public static List<SignalAccountRecord.PinnedConversation> localToRemotePinnedConversations(@NonNull List<RecipientRecord> settings) {
    return Stream.of(settings)
                 .filter(s -> s.getGroupType() == RecipientTable.GroupType.SIGNAL_V1 ||
                              s.getGroupType() == RecipientTable.GroupType.SIGNAL_V2 ||
                              s.getRegistered() == RecipientTable.RegisteredState.REGISTERED)
                 .map(StorageSyncModels::localToRemotePinnedConversation)
                 .toList();
  }
```
This is the method that creates the list of pinned conversations for sync. It explicitly checks if the conversation is either a group conversation, or a registered contact. This method is used in both a linked device, and in primary.

The pinned contacts list is synched alongside things like the profile name, and link preview settings. In a worst-case scenario, it could be synched on every received message. Even if the above filter was removed from the linked device, it would only remove the unpinning in one direction. Eventually, the primary would send a "This user is unregistered, therefore unpin it" sync to the linked device, which would then send "This user is unpinned for me, unpin it" back to the primary.

In other words, if this is a use case for you, it needs to be fixed upstream. And, considering it was done intentionally, it would probably be an uphill battle to get it changed.

johanw666:
@SlugFiller yes, I just found that code myself. No problem, since I self compile with some changes anyway I can just remove that filter line.

BTW, I accidentily replaced Molly 6.9.2-linkdevice with Molly 6.11.7 and it keeps receiving messages for the linked device and it sais it is registered with the same phone number. Is this expected? When I write something in note to self it gets synced, when I send a message from another device both installs receive it and read status gets synced too.

SlugFiller:
@johanw666 The keys and device id are stored the same on a linked and primary device, so even if you change clients after link, the core functionality would still work. The initial version of this PR didn't even make any changes to the core functionality, only the registration flow. However, currently there are a few differences:

1. Without this PR, a linked device believes it lacks a pin (even though linked devices have no pin), and cannot detect it received a pin from the primary. It will occasionally bug you to give it a pin.
2. Various options that can only be modified via the primary will appear in the settings. In a "true" linked device, they would be greyed out, with a message saying "This setting can only be changed via the primary".
3. If profile synchronization hasn't yet finished, it may either request you to input new profile details, or force-push an empty profile to the server. However, to reach this condition, you'd have to replace the client immediately after linking, as the profile would only need a couple of seconds to update anyway.
4. If you change the phone number from the primary, the fake-linked device would become unlinked. However, since nobody bothered testing this feature, I can't guarantee that the real linked device would behave any differently.
5. Pin storage synching would not work. The linked device would simply detect that it has no pin, and skip it. Ironically, this would actually solve your issue with pinned conversations.
6. The list of linked devices would be a bit less friendly, and may allow a fake-linked device to unlink itself. Although, since this was never tested, the results of trying to do that are anyone's guess.

SlugFiller:
A note about the new merge: Since Signal added a new "reporting token" item (used by the new spam report feature) to the serialized envelope, I've decided to push the "updated pni" token one index back, to avoid conflict. This makes it easier to upgrade from "clean" Signal/Molly to this PR, at the cost of upgrading from a previous version of this PR to the current. However, the issue would only manifest if there was a pending phone number change message that wasn't yet processed by the device. Since nobody seems to be willing to test a phone number change, this is unlikely to be an issue anyone encounters. However, moving forward, this may become a recurring issue every time Signal add a new data point to the envelope.

A possible solution is to set the "updated pni" index very high, to avoid conflicts long term. The down side would be that this would reduce the odds of a clean upstream merge, since upstream would probably prefer the indices to be sequential. I'm also not too familiar with how protobuf handles a large index gap. Presumably, it just makes a sparse serialization. But, if it causes space waste, or worse, an error, that would be less than desirable.

valldrac:
> Since Signal added a new "reporting token" item (used by the new spam report feature) to the serialized envelope, I've decided to push the "updated pni" token one index back, to avoid conflict.

Where is this serialized envelope used? You say later that it will only cause a problem if there is a "pending phone number change message" but I'd like to fully understand this. My first thought is to use a higher ID to follow protobuf rules. Using 2000 as starting index will only increase the serialized message by one byte.

SlugFiller:
@valldrac 
> Where is this serialized envelope used?

Signal Android serializes all received packets prior to processing, so that in the event that the app is killed before the packet can be processed, it can resume processing when the app is re-launched, without having to re-download any packets. This is related to the jobs system, since packet processing is done in a job, and jobs are also persisted (as those also need to survive the app being killed).

The serialization format used internally is different from the one used by the network protocol. As in, the message is already partially unpacked when it is serialized. The (original) network protocol has the updated PNI as a reserved index, but the processing queue serialization discards the value completely, so I had to add an index for it, so it can be read in `handleSynchronizePniChangeNumber`.

The "updated PNI" is part of the phone number change protocol. PNI, as I understand it, stands for Phone Number Identity, and is a set of cryptographic keys that represent a confirmation that a certain device passed SMS verification. All linked devices (and primary) must have the same PNI. They initially receive it during linking. However, if the primary changes phone numbers, then it must also change PNI. As part of the phone number change protocol, the primary has to craft a packet for each linked device, giving them updated credentials, and attach it to the phone change code verification request.

The credentials are received by the linked device as a sync response of type "PniChangeNumber". However, for some reason, the updated PNI is present in the envelope instead of the sync contents. Presumably because it is needed in order to decrypt the rest of the message, or something along those lines.

I've pushed a version with `updatedPni` index set to 2000 for both the envelope and metadata (Metadata is used for moving data from the envelope to the content, so it can be processed in `MessageContentProcessor`. Don't ask me why this is so convoluted), so now people can test and see if it crashes horribly. Assuming someone actually bothers testing number change.

johanw666:
I removed the linkage, the linked Molly shows a "Device no longer registered" banner as expected. Now I can't find a way to link it again, unless I remove all data first and restore from backup the existing messages. Is this workflow not (yet?) implemented?

SlugFiller:
@johanw666 If you tap on the banner, it should lead you back to the registration process, and you can click on the link device button again. I actually got to test that flow a little while ago, and it seemed to work as expected.

johanw666:
Ah, I assume something must be wrong in the UI: I see some fragments of a text that is falling away at the bottom of the screen.

I run it on a Samsung A50, screen is 2340 x 1080.

Screenshot_20230223-015349_Molly

SlugFiller:
@johanw666 Yes, that text is "Link to existing device". The "Continue" itself should not be this low. Also, the same UI is used for re-registration as for the first registration, so assuming you managed to link the device once, this issue was not present the first time. Presumably, a change was made to the layout in the latest version. I would have to look into it. It probably requires some change in `app/src/main/res/layout/fragment_registration_enter_phone_number.xml`, but debugging UI layout is a challenge

jaj42:
I downloaded the latest pre-built version v6.11.7-1+linkdevice.rev7 and I don't see any link at all, there is only the Continue button

jaj42:
I downloaded the latest pre-built version v6.11.7-1+linkdevice.rev7 and I don't see a link at all, only the Continue button. This is on MIUI with the FOSS version. I can't take a screenshot because the app prohibits.

johanw666:
> I can't take a screenshot because the app prohibits.

You can allow screenshots in Settings - Privacy - Screen security.

jaj42:
> > I can't take a screenshot because the app prohibits.
> 
> You can allow screenshots in Settings - Privacy - Screen security.

I don't think these settings are available before you register

SlugFiller:
Okay, I managed to fix the link screen layout, to the best of my efforts. Had some issues because the resource minification makes it hard to rebuild, and also seems to break debug builds for some reason. But, at any rate, it works again, now.

johanw666:
Thanks, this does work. Re-registration succeeded, although I was stilll stuck in Molly when I switched to Signal on the same device to read the QR code. But when I closed Molly by using the back button several times and reopened it linking was OK.

johanw666:
I'm still getting some crashes, one backup related and the other looks like missing keys or something: https://debuglogs.org/android/6.11.7-1.0-JW-linkdevice/8d03720d5431002601bfc390d388be80faa051041067862a4e4738e98f2ed25c

SlugFiller:
@johanw666 The export issue seems to stem from a file in the avatar directory having a name that can't be parsed as a recipient name. You mentioned building your own client. If possible, try to build a version that logs the file list in `app/src/main/java/org/thoughtcrime/securesms/profiles/AvatarHelper.java::getAvatars`.

The most naive assumption would be that the avatar for the user is saved in a bad name, but this seems unlikely since the user has a valid recipient id, and also, the avatar is downloaded on a non-linked device as well. It would also assume you're the only person who tried the client and has periodic backup enabled.

The more likely alternative is that your avatar folder was just corrupted somehow, or had an unwanted file placed in it by some system utility. But without knowing the name of the file causing the issue, it's hard to make any guesses.

The most extreme option is that it's failing on something like "..". But if that was the issue, it would be present on the main device as well, so long as it had periodic backups active.

The debug log is pretty long, and I couldn't find the missing keys issue. I did see another issue related to the exporter, dealing with reading from the database, which farther suggests some sort of corrupted data, this time in the message table. Dumping the database would be harder than dumping a file list, and wouldn't be recommended anyway, for security reasons.

There were a few warnings about received recipient id for self. As far as I can tell, these are normal, and occur after re-linking a device. It receives queued messages intended for the previous linked keys, which the new device can't decrypt. After the queue is cleared, these messages stop.

johanw666:
Sorry to reply so late, I was sick. I've added logging but I don't see anything logged after a crash:

```
  public static Iterable<Avatar> getAvatars(@NonNull Context context) {
    File   avatarDirectory = getAvatarDirectory(context);
    File[] results         = avatarDirectory.listFiles();

    if (results == null) {
      // JW: extra logging
      Log.i(TAG, "getAvatars: nothing found");
      return Collections.emptyList();
    }

    return () -> {
      return new Iterator<Avatar>() {
        int i = 0;
        @Override
        public boolean hasNext() {
          return i < results.length;
        }

        @Override
        public Avatar next() {
          File file = results[i];
          try {
            // JW: extra logging
            Log.i(TAG, "getAvatars: " + file.getName());
            return new Avatar(getAvatar(context, RecipientId.from(file.getName())),
                              file.getName(),
                              ModernEncryptingPartOutputStream.getPlaintextLength(file.length()));
          } catch (IOException e) {
            return null;
          } finally {
            i++;
          }
        }
      };
    };
  }
```

SlugFiller:
@johanw666 The crashlog you gave definitely went over the lines above, so if it's the same crash/crashing in the same place, it would have to trip over those log lines.

If it's not, either you're getting a different crash, or something went wrong during compile/deploy.

johanw666:
I had to reinstall Signal because messaging failed on that device after I upgraded to 6.19.6. (it works again now).

However, while re-linking 2 desktop instances was no problem I kept getting "network error" when I try to re-link Molly linkdevice.

Debug log: https://debuglogs.org/android/6.18.4-1.0-JW-linkdevice/d9eb20b408990fe1b33bf0e276261c0aae739f5d3d5ed5b8d7085a192b79996e

akki42:
Hi @SlugFiller ,
just wanted to let you know that I have been using Molly-Insider-v6.20.5-1+linkdevice.rev8.apk for about two weeks now and have had no problems whatsoever yet. Great work!
(Linked device: Samsung Tab S5e, LineageOS 20)
Is there anything that you would like me to test? Would love to see this in main...

johanw666:
@SlugFiller any time schedule on when the update to Molly 6.23 can be expected?

SlugFiller:
The biggest change in this update is that the `PNI_IDENTITY` sync request was removed. As in, handling the request was removed from the primary, so a linked device can no longer make the request. From what I can tell, PNI only changes when the phone number is changed or re-registered, so presumably the existence of a dedicated change number process renders it redundant.

I still haven't received a single test report for the aspects described [here](https://github.com/mollyim/mollyim-android/pull/157#issuecomment-1371215644). Should I assume there's a lack of interest? For the past few months the only changes to this was just repeated rebasing.

p1gp1g:
Small bug (?) : I need to force stop molly and restart it to load the list of groups I'm member of

valldrac:
A Matrix user gave us this feedback.

- Deleting messages does not work.
- Trying to link across multiple user profiles is unreliable.
- Subsequent linking of devices after unlinking the first time is broken, you need to reinstall the app or clear its storage.
- Linking devices clears pinned chats on the host.
- The linked device asks for you to make a pin for some reason.
- Less important but in the subsequent sign up menu, the cancel button and the link device button appear almost overlapping, both work though.

alfred6427:
> Does anything in Settings->Account work from a linked device?

- Create a PIN and other PIN settings  **Does not seem to work**
- Change phone number (theoretically should not work)  **Did not test**
- Transfer account  **Works, doesn't log you out of the devices till you finish activating.**
- Delete account **Works.  Deletes the Signal account**

Does changing numbers from main work? Does the linked device continue to work, or gets unlinked? -- **Did not test**

Some other observations

-  Creating a PIN from a linked device does not appear to do anything
-  Activating Registration Lock crashes the app most of the time.
-  Enable/Disable PIN Registration doesn't appear to work.  When switching it off on linked device, it doesn't change on the Primary device
- Unlinking works fine from Primary device


Deleting messages from a linked device seems to work in a group, but not on the note to self group. The other device recieves a notification saying "[Internal-only] Recieved an invalid message!".  I don't have an extra number to test with, so not sure if it works differently with multiple accounts in a group.

Notes on linking a device in a secondary Android User Profile.  **These issues are not present when using a physical secondary device.**

- Linking across Android User profiles is not very consistent.  Sometimes doesn't work. 
- Prompts you to create a PIN.  Force closing the app and relaunching fixes the problem.  
- When relinking, you have to clear cache and storage.

akki42:
> A Matrix user gave us this feedback.
> 
>     * Deleting messages does not work.

Just wanted to mention: this "does not work" with / from a linked Signal Desktop (Linux / Flathub) either - so this seems to be "by design".

johanw666:
Indeed, if you delete locally it is only deleted from the device you delete it on, not from linked devices. That has always been so.

nospace-here:
Compared to Molly on the phone, I don't see the navbar to switch between chats and call log on the tablet. Will this be re-added later or might the design slightly change?

nospace-here:
Recieved an error on the linked devices when I sent something from the main molly device and seconds after delete it for all. On the main Molly device it's deleted, but not on Signal Desktop or the linked molly device with the latter giving an error notification https://debuglogs.org/android/6.23.5/a71522836c32ab4506833646851f9d990da5ae576fdb6e4535df23596ee7f114

njmdietrich:
Is this still in active development? I am using the latest apk with this on my tablet, but it is behind by a few versions now. Would love to see either an updated build or for this to be merged.

xyz-nobody:
yes same i use old version and its very usefull this feature


------- Original Message -------

nospace-here:
It's likely that Signal is now working on tablet support, since they'd get ranked lower on Googleplay if they don't https://community.signalusers.org/t/android-version-of-signal-now-available-on-tablets-and-chromeos/54570/7

valldrac:
Hey @SlugFiller, let's see if you can pick up this pull request again. Honestly, it's generated a lot of interest and I'd love to get it merged as soon as possible.

In the meantime, I've gone ahead and rebased to main. I also fixed a bug related to uploading pre-signed keys at registration, due to Signal's recent updates. You can find my branch [here](https://github.com/mollyim/mollyim-insider-android/tree/linkdevice).

And for those of you who want to test and use this feature, there's a Insider release ready for download, based on v6.30.4-1:

https://github.com/mollyim/mollyim-insider-android/releases

xyz-nobody:
Thank you for this update, I have manually updated to v6.30.4-1+linkdevice.rev10 and so far I haven't noticed any bugs.
Tomorrow I'll continue testing


------- Original Message -------

SlugFiller:
- Rebased
- Disabled settings confirmed to not work
- Applied aforementioned fix to key upload
- Shamelessly stole prettier use of parameters and imports
- Moved around elements to make the registration dialog look right again.

Regarding secondary android profile: Linking requires both primary and linked to be active and in the foreground in order to complete. I'm not too familiar with android multi-profile, but assuming at least one of the profile needs to be in the background, it is surprising that it works ever.

Regarding pinned chats: I have difficulty testing it, since synchronizing those depends on storage sync, which doesn't work with a signal-cli main. I'll try to dig into it, but I want to know one thing first: Does it work as expected with Signal Desktop?

Regarding relinking: Tested and works fine for me. Without so much as restarting the app. Certainly without clearing cache. But re-testing the link process allowed me to do some adjustments to the UI.

@valldrac Looking at your rebase, it seems like you have a pretty good grasp of the code of this PR already. The differences from mine are mostly cosmetic, and it even helped me get mine done faster and/or better in some points.

If you have a better setup to debug the pinned chats thing, you can probably take this over already.

valldrac:
Hey, thanks for the detailed update and the hard work you've put into this. I'll look into the pinned chat issue as you suggested, but everything else looks good to me. I'm going to merge it.

Getting all the pieces to fit together to make this feature come to life was challenging, so kudos for handling it. I'm sure the users will love this new feature. Great job!

akki42:
Hi all,
First: many thanks for getting this feature into Molly!

I've been using Molly-Insider-v6.30.4-1+linkdevice.rev10 (and previous branch releases) over the last weeks. Today, I've activated Chat - Backups (in an attempt to transfer the data to newly released Molly).
Unfortunately, the backups fail (at 99%) and Molly Insider force-restarts.
(After deactivating Backups, everything works fine again.)

I don't know, whether this is related to the linkdevice functionality or a more general Molly issue.

Log file at https://debuglogs.org/android/6.30.4/1ca7c546ef10d2511bd56aa49248a1a42e55719e6bfd5997d734e4ce270e11d5.

njmdietrich:
I can confirm that backups fail with the latest apk from here. However, they work fine on the newest molly release on a linked device. Since I was missing old messages on the linked device anyway, I didn't mind losing the linked device's database and just started anew.

akki42:
Thanks, @njmdietrich, for testing (so fast)!

And it's good to know that "it" (whatever the cause) has already been fixed with the official Molly release.

Now that leads me to a more general question:
Is it possible to import existing messages / backups into a (to be) linked device?
(I would of course love to have all my messages on the linked device as well. So if I could import from the main device somehow, that would be great. As far as I know, that's not possible with the official Signal (desktop or iOS) apps - but maybe Molly can improve oin that?)

njmdietrich:
@akki42 your suggestion intrigued me so I tested it. I was able to import a backup from my main device (which runs signal-foss) and link the device just fine, but afterwards the link seems kind of broken. I can send and receive messages from the linked device just fine, but sent messages from my main device don't sync with the linked device. I guess I will open a new issue about this

akki42:
Sorry for getting off-topic:
In view of your, @njmdietrich, promising report, I tried to import my main devices history into a newly installed molly on a secondary (to be linked) device.
Unfortunately, I made some foolish - in hindsight - attempts and now ended up completely locked out of my signal account on all devices due to "registration lock" being activated. 
As I still have / know
- my original Signal app (registration lock) PIN
- access to the registered phone number
- this morning's backup file and
- the previous "main" device (but with a fresh signal installation, so no user data)
could anyone please let me know the best way to "restore" (my main device) as much as possible?
(If I wait for seven days - bad enough, but manageable - and then re-register, will I be able to reimport my old messages? Will my contacts and groups still "recognize" me?)

akki42:
@njmdietrich 
I can finally confirm: importing my main device's (Signal from Play Store) backup to my secondary device (current Molly v6.34.5-1) and linking the secondary device worked just fine today. (I couldn't test earlier because main device's Signal had higher version than Molly, which would not import such backup files...)

But:
Unfortunately, backup on the secondary device [again](https://github.com/mollyim/mollyim-android/pull/157#issuecomment-1719915933) fails and forces Molly restart (despite newest Molly version).

Log file looks very similar to the one posted two weeks ago:
```
09-30 09:58:20.590 32362 32414 D FullBackupExporter: Exporting table: payments
09-30 09:58:20.592 32362 32414 D FullBackupExporter: Exporting table: pending_pni_signature_message
09-30 09:58:20.593 32362 32414 D FullBackupExporter: Exporting table: reaction
09-30 09:58:20.599 32362 32414 D FullBackupExporter: Exporting table: remapped_recipients
09-30 09:58:20.600 32362 32414 D FullBackupExporter: Exporting table: remapped_threads
09-30 09:58:20.601 32362 32414 D FullBackupExporter: Exporting table: remote_megaphone
09-30 09:58:20.604 32362 32414 D FullBackupExporter: Exporting table: sticker
09-30 09:58:21.962 32362 32414 D FullBackupExporter: Exporting table: storage_key
09-30 09:58:21.965 32362 32414 D FullBackupExporter: Exporting table: story_sends
09-30 09:58:22.510 32362 32414 W LocalBackupJobApi29: Backup failed. Deleted temp file
09-30 09:58:22.512 32362 32414 I GenericForegroundServic: Stopping foreground service id=0
09-30 09:58:22.515 32362 32362 I GenericForegroundServic: handleStop()
09-30 09:58:22.516 32362 32362 I GenericForegroundServic: Last request. Ending foreground service.
--------- beginning of crash
09-30 09:58:22.517 32362 32414 E AndroidRuntime: FATAL EXCEPTION: signal-JobRunner-3
09-30 09:58:22.517 32362 32414 E AndroidRuntime: Process: im.molly.app, PID: 32362
09-30 09:58:22.517 32362 32414 E AndroidRuntime: org.thoughtcrime.securesms.recipients.RecipientId$InvalidStringRecipientIdError
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.recipients.RecipientId.from(RecipientId.java:52)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.profiles.AvatarHelper$1.next(AvatarHelper.java:78)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.profiles.AvatarHelper$1.next(AvatarHelper.java:67)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.backup.FullBackupExporter.internalExport(FullBackupExporter.java:190)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.backup.FullBackupExporter.export(FullBackupExporter.java:119)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.jobs.LocalBackupJobApi29.onRun(LocalBackupJobApi29.java:143)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.jobs.BaseJob.run(BaseJob.java:31)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.jobmanager.JobRunner.run(JobRunner.java:86)
09-30 09:58:22.517 32362 32414 E AndroidRuntime: 	at org.thoughtcrime.securesms.jobmanager.JobRunner.run(JobRunner.java:49)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: org.thoughtcrime.securesms.recipients.RecipientId$InvalidStringRecipientIdError
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.recipients.RecipientId.from(RecipientId.java:52)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.profiles.AvatarHelper$1.next(AvatarHelper.java:78)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.profiles.AvatarHelper$1.next(AvatarHelper.java:67)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.backup.FullBackupExporter.internalExport(FullBackupExporter.java:190)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.backup.FullBackupExporter.export(FullBackupExporter.java:119)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.jobs.LocalBackupJobApi29.onRun(LocalBackupJobApi29.java:143)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.jobs.BaseJob.run(BaseJob.java:31)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.jobmanager.JobRunner.run(JobRunner.java:86)
09-30 09:58:22.517 32362 32414 E SignalUncaughtException: 	at org.thoughtcrime.securesms.jobmanager.JobRunner.run(JobRunner.java:49)
09-30 09:58:22.533 32362 32414 I JobManager: Successfully flushed.
```

Are backups still working for you on linked device?

Many thanks,
akki42

valldrac:
We're tracking the backup crash in #219.

obacht8:
Thank You so much for that "linked device"-feature !!!

PR #170, opened 2023-03-06
Allow users to backup and restore without having to reverify

dylangerdaly:
Backup all key values


johanw666:
Interesting. But wouldn't it be better to make this an option that can be put on and off? If you restore this on a another device and the original device is switched on as well, funny things might happen.

dylangerdaly:
Absolutely, I don't know how to write code for ratio buttons etc, if anyone can help that'd be great.

I imagine it'd be better to have an option to unset one of the SharedPrefs on the old device.

Ultimately it should be gated behind an option and be properly tested etc

-------- Mensagem original --------
Em 7/03/2023, 17:47, johanw666 escreveu:

johanw666:
I am considering copying this to my own fork. Adding an option is easy: put all the code you added between
```
if (option) {

}
```
blocks. Molly prefers using options in keystore AFAIK, I usually use SharedPrefs bt the effect would be the same.

Then add the button in the xml file of the control file you want to edit and the event handler in the Kotlin code (probably in components/settings/app/chats), just look how the other controls are made and copy that.

I'll try to write that code soon, did you test if this works as stated? I don't see anything wrong but you never know.

dylangerdaly:
> I'll try to write that code soon

Cheers

Yeah it works fine for me, tested it a couple times

valldrac:
In my opinion, the best way to handle this situation is by using the device transfer function to make an exact duplicate of the app instead of relying on backups. By doing this, you can deactivate the original app during the migration. It's not a good idea to include these values in backups because it will lead to multiple issues.

dylangerdaly:
No worries, this was just more to show it's possible.

> by using the device transfer function

I haven't used this method of migration before, how is it different to backups?

> By doing this, you can deactivate the original app during the migration.

You can do this now too, you just flip `PROMPTED_PUSH_REGISTRATION_PREF` OFF/false and the old Molly will think it's in a brand new state, kinda like flipping the light switch off on your way out.

valldrac:
I'm going to close this pull request so it doesn't stay open forever. We've got to fix this issue, but we'll address it in a different way.

Issue #14, opened 2020-10-12
Use OpenStreetMap as location provider


valldrac:

iiitomo:
Probably code from signal-foss can be reused , see https://forum.f-droid.org/t/ive-degoogled-signal-messenger/10443/76

valldrac:
Seems it was too early to set it as done. There are some issues to work on before it's ready to release.

- Move tile database to cachedir
- Encrypt tile database
- Set proxy settings
- Check for DNS leaks when the proxy is configured

I've moved the development of this feature to an experimental branch.

carallo:
Is it possible to use OSM as location provider but still using Google's FCM as push notification provider? I am on microg and have issues with Google Maps provider, but still need FCM due to battery draining...

BTW, thank you so much for your work!

valldrac:
Yes. Google Maps is going to be replaced by OSM in both Molly and Molly-FOSS. Remember that the standard version Molly already uses FCM.

There are some issues we have to fix first. See my comment above.

carallo:
> Yes. Google Maps is going to be replaced by OSM in both Molly and Molly-FOSS. Remember that the standard version Molly already uses FCM.
> 
> There are some issues we have to fix first. See my comment above.

Perfect, thank you so much! 
I know standard Molly version uses FCM, but being on microg location sharing does not work with Google Maps provider, so I wanted to try OSM version to see if it fixes the issue, but I still need FCM.
I will be waiting till OSM gets implemented, thank you again and good luck!

ronidee:
Can the OSM lib or API used as a drop-in replacement for the Google version? 
At least moxie said he'd be (or was..) open to merge a PR if this would work, iirc.

MrCybertux:
@valldrac i do not know if this helps you but the guys over at telegram FOSS have a working implementation of OSM
https://github.com/Telegram-FOSS-Team/Telegram-FOSS

baarkerlounger:
@valldrac the readme still refers to Molly-Foss not supporting a location provider. Is it correct to say it now does?

valldrac:
> @valldrac the readme still refers to Molly-Foss not supporting a location provider. Is it correct to say it now does?

Nice catch! I need to update the README.

PR #177, opened 2023-04-15
Further secure the osmdroid implementation

h3ph4est7s:
Update osmdroid version and add further security controls to the initial implementation #14 
- [x] Move tile database to cachedir
- [ ] Encrypt tile database
- [x] Set proxy settings
- [x] Check for DNS leaks when the proxy is configured
- [x] Update osmdroid library to 6.1.16
- [x] Add SSL pinning  
- [ ] Disable HTTPS caching (needs modification of the library)


app/src/main/java/org/thoughtcrime/securesms/net/SocksProxy.java
```diff
@@ -49,7 +49,7 @@ public String getUrl() {
   public InetSocketAddress getSocketAddress() {
     InetSocketAddress address = null;
     try {
-      address = new InetSocketAddress(host, port);
+      address = InetSocketAddress.createUnresolved(host, port);
```
valldrac, line 52: I'm not sure, but I'd say the address needs to be resolved at this point. Could you try setting a non-trivial hostname (i.e. fully qualified DNS name, not localhost) and test if it still works?

app/src/foss/java/com/google/android/gms/maps/MapView.java
```diff
@@ -53,8 +58,14 @@ public void getMapAsync(OnMapReadyCallback callback) {
...
-    Configuration.getInstance().load(context, PreferenceManager.getDefaultSharedPreferences(context));
-    Configuration.getInstance().setUserAgentValue(userAgent);
+    final IConfigurationProvider config = Configuration.getInstance();
+    final NetworkManager         nm     = ApplicationDependencies.getNetworkManager();
+    config.setOsmdroidTileCache(new File(context.getCacheDir(), "tiles"));
+    config.load(context, PreferenceManager.getDefaultSharedPreferences(context));
+    config.setUserAgentValue(userAgent);
+    if(nm.isProxyEnabled()) {
+      config.setHttpProxy(nm.getExistingProxy().makeProxy());
+    }
```
valldrac, line 68: Don't use the NetworkManager to get the proxy settings. Use this instead:

```java
import org.thoughtcrime.securesms.net.Network;

Proxy proxy = Network.getProxy();
if (proxy != null) {
    // ...
}
```

app/src/main/java/org/thoughtcrime/securesms/net/NetworkManager.java
```diff
@@ -104,6 +104,8 @@ public int getDefaultProxySocksPort() {
     return OrbotHelper.DEFAULT_PROXY_SOCKS_PORT;
   }
 
+  public SocksProxy getExistingProxy() { return isProxyEnabled() ? existingProxy : null; }
```
valldrac, line 107: This function can be removed. `Network.kt` provides the proxy object.

app/src/foss/java/com/google/android/gms/maps/MapView.java
```diff
@@ -53,8 +58,14 @@ public void getMapAsync(OnMapReadyCallback callback) {
 
   private void setDefaultConfiguration(Context context) {
     String userAgent = String.format("%s/%s", context.getString(R.string.app_name), BuildConfig.VERSION_NAME);
```
valldrac, line 60: Set the user agent to simply ```context.getPackageName()```.

This was done by the original author of the PR, but better to not leak the version and title of the app.

app/src/foss/java/com/google/android/gms/maps/MapView.java
```diff
@@ -53,8 +58,14 @@ public void getMapAsync(OnMapReadyCallback callback) {
...
-    Configuration.getInstance().load(context, PreferenceManager.getDefaultSharedPreferences(context));
-    Configuration.getInstance().setUserAgentValue(userAgent);
+    final IConfigurationProvider config = Configuration.getInstance();
+    final NetworkManager         nm     = ApplicationDependencies.getNetworkManager();
+    config.setOsmdroidTileCache(new File(context.getCacheDir(), "tiles"));
+    config.load(context, PreferenceManager.getDefaultSharedPreferences(context));
```
valldrac, line 64: Another vestige of the original PR. Molly doesn't persist any preference value for OSM. We can safety remove this call to `load()` and rely on the default values provided by the library.

valldrac (changes requested):

app/src/main/java/org/thoughtcrime/securesms/net/SocksProxy.java
valldrac: Tested and it works! I also fixes a potential ANR error.

valldrac (approved):

PR #183, opened 2023-04-26
Update Add more AMOLED Dark.xml

JohnBetaro:
One of the cool things I like about Molly is the black theme, but I hope you'll accept this change to make it even cooler.

### DIFF
--- a/app/src/main/res/values/material3_colors_dark.xml
+++ b/app/src/main/res/values/material3_colors_dark.xml
@@ -4,8 +4,8 @@
     <color name="signal_dark_colorPrimaryContainer">#4A485E</color>
     <color name="signal_dark_colorSecondary">#C8C5E0</color>
     <color name="signal_dark_colorSecondaryContainer">#46445C</color>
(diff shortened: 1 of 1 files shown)


BHydden:
Why did you close https://github.com/mollyim/mollyim-android/pull/182 ?

JohnBetaro:
Just by mistake

valldrac:
Thanks for the pull-request! Can you add some screenshots so we can see the difference?

JohnBetaro:
As per your request, I hope you like it.

p1gp1g:
It would be nice to have a before after to see the changes. And having the related variable, if we want to keep/ change some.

You can use a table to do this
```
Before | After | Vars
---|---|---
Img1 | Img2 | signal_dark_color

p1gp1g:
I'm OK that PR currently reduces accessibility. That's why I've suggested to use a darker color than current ones but not full black.

I'm also OK that AMOLED black may reduces accessibility. I think the best thing would be having 3 themes : light, dark, black

PR #184, opened 2023-04-26
Add position indicator to OSM

p1gp1g:
This adds the position indicator on OSM.


valldrac:
I went with merging #185 instead because it seems like sharing a location through Google Maps' URLs is the standard way to do it.

PR #185, opened 2023-04-26
Always use Google maps to share position

p1gp1g:
Gmaps links are unfortunately like an universal geo: URI, all platforms use these links to open their own GPS app (iOS/Android/Browsers). Even OSM apps on Android use it.

This solution is probably better than #184 since it makes position sharing more usable for everyone.

Another solution would be to have the 2 links (OSM and Gmap) but it is against signal UX, keeping things easy. Then users would have to say 'click on the second link'.


nisc:
It seems that in the current version, both links are always included.

Personally I'd prefer to have only the Google link. I agree with OP's statement.

> Gmaps links are unfortunately like an universal geo: URI, all platforms use these links to open their own GPS app (iOS/Android/Browsers). Even OSM apps on Android use it.

PR #238, opened 2023-11-11
Display MayHaveMessagesNotification on push with locked db

p1gp1g:
This works with FCM : If the remote message's priority is high, then the push notification should require a notification :)

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/gcm/FcmReceiveService.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/gcm/FcmReceiveService.java
@@ -18,6 +18,7 @@ import org.thoughtcrime.securesms.registration.PushChallengeRequest;
 import org.thoughtcrime.securesms.service.KeyCachingService;
 import org.thoughtcrime.securesms.util.NetworkUtil;
 import org.thoughtcrime.securesms.util.SignalLocalMetrics;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
 
 import java.util.Locale;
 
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/thoughtcrime/securesms/gcm/FcmReceiveService.java
```diff
@@ -28,6 +28,10 @@ public class FcmReceiveService extends FirebaseMessagingService {
   @Override
   public void onMessageReceived(RemoteMessage remoteMessage) {
     if (KeyCachingService.isLocked()) {
+      if (remoteMessage != null && remoteMessage.getPriority() == RemoteMessage.PRIORITY_HIGH) {
+        Log.d(TAG, "New urgent message received while the db is locked");
```
pixincreate, line 32:
```suggestion
        Log.d(TAG, "New high priority message received while the database is locked");
```

pixincreate (commented):
Other than that, looks good to me.

Just a quick question, this implementation only works for FCM and not UnifiedPush and websockets, right?

p1gp1g:
Thanks for you review. For information, urgent is [a signal term](https://github.com/mollyim/mollyim-android/blob/02fa8abb4b0683a454c8314a9a9a87f30670a4da/libsignal-service/src/main/protowire/SignalService.proto#L36) which is [converted by the server to FCM term](https://github.com/signalapp/Signal-Server/blob/38bc0c466a263999751538a7a1eb2231ea459956/service/src/main/java/org/whispersystems/textsecuregcm/push/FcmSender.java#L87).

> Just a quick question, this implementation only works for FCM and not UnifiedPush and websockets, right?

I've done this for UnifiedPush too : https://github.com/mollyim/mollyim-android-unifiedpush/commit/9f9d882035f5dd8506c309bad042c3d995da71a9 but it requires the version 1.1.0 of mollysocket I will publish during the day

PR #251, opened 2023-12-11
Update README.md

phoropter:
Add text denoting mollysocket requirement for Molly-UP, unavailability for linked devices, and updated table.


BHydden:
In line 73 it might be good to directly state whether Molly-UP has blobs or not. The way it is written in your proposal I think leaves it ambiguous.

BHydden:
Yeah sounds good thanks for that.

phoropter:
I think with 3 choices, a bullet list makes readability easier.

BHydden:
I agree. Latest version is very clear ❤️

valldrac:
Thank you for the contribution. I've corrected some errors and made some changes to enhance clarity.

Issue #453, opened 2025-03-12
Change navigation bar chats icon to Molly's icon


Turhvjbufv:
### Guidelines

- [x] Yes, I have searched the existing requests

### Feature Description

Currently the chat icon inside the app looks like this 
![Image](https://github.com/user-attachments/assets/dd50aa85-a782-477f-80cf-61063e46f56d)
to change it to the molly icon (basically add the eight circles to it) 
Why? Makes it look less plain and more intriguing

valldrac:
Thanks for the suggestion. We agree, but that icon (and its animation) is in Lottie format, so it's not easy to change:

https://github.com/mollyim/mollyim-android/blob/main/app/src/main/res/raw/chats_28.json

We'll keep it noted for when we can dedicate time to it.

ex-alice:
All three buttons in active state are solid (full) which looks really good. If you add eight circles and button is in active state it will lose its magic. In my opinion there's **no need** to add anything or change it. Currently it **looks good** as it is.

Yolakalemowa:
@valldrac hi, I tried asking a question in a new issue but it won't direct me to the matrix room. Can I ask you where did the Weather icon go after most recent update? It's replaced with a moon icon :( 

I loved that weather icon.

valldrac:
After discussing it with our designer, we would like to maintain the solid look, but change the shape to match Molly's logo.

PR #457, opened 2025-03-13
Changed navigation bar chats icon to Molly's icon #453

Misenpai:


valldrac:
Thank you, but we would like to maintain the solid look of the icon. What needs to be done is to change the shape so that it resembles Molly's logo.

Issue #391, opened 2024-11-19
Better censorship circumvention 


UjuiUjuMandan:
### Is there an existing request for this?

- [X] I have searched the existing issues

### Feature description

As I've mentioned in the pull request ignored by Signal https://github.com/signalapp/Signal-Android/pull/13681, 2 out of 3 of the current Fastly CDN fronting URLs are blocked by CCP. Rather than changing the fronting URLs to other static value, we better to let it, the censorship circumvention to be customizable, by users rather than the Signal developers. For example:

- A switch for selecting domain fronting with Google or Fastly.
- An input for replacing default fUrl and GHost.

PR #534, opened 2025-07-03
Update fConfig fronting domains for censorship circumvention

UjuiUjuMandan:
Initial pull request:

* https://github.com/signalapp/Signal-Android/pull/13681

Former request:

* https://github.com/mollyim/mollyim-android/issues/391

I still think that the domains of social networking sites are more likely to be blocked and should be replaced to something more solid.

This patch replaced the Reddit and Pinterest domains with `www.gov.uk`, which also uses Fastly CDN, as they [stated on their website](https://docs.publishing.service.gov.uk/manual/cdn.html) and [confirmed](https://dnschecker.org/#A/www.gov.uk). The censors of Iran, Cuba, and China will face heavy collateral damage blocking it.

My former pull request to Signal also included `www.python.org`, this programming language is popular in China, and it might also be in Iran. Cuba is unknown.

* https://www.iranintl.com/en/202304146827 "‘Python-Gate’: Raisi Administration’s New Embarrassment"

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/push/SignalServiceNetworkAccess.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/push/SignalServiceNetworkAccess.kt
@@ -74,8 +74,6 @@ class SignalServiceNetworkAccess(context: Context) {
     private const val HTTPS_CLIENTS_4_GOOGLE_COM = "https://clients4.google.com"
     private const val HTTPS_GOOGLEMAIL_COM = "https://googlemail.com"
     private const val HTTPS_GITHUB_GITHUBASSETS_COM = "https://github.githubassets.com"
-    private const val HTTPS_PINTEREST_COM = "https://pinterest.com"
-    private const val HTTPS_WWW_REDDITSTATIC_COM = "https://www.redditstatic.com"
     private const val HTTPS_WWW_GOOGLE_COM_EG = "https://www.google.com.eg"
     private const val HTTPS_WWW_GOOGLE_AE = "https://www.google.ae"
     private const val HTTPS_WWW_GOOGLE_COM_OM = "https://www.google.com.om"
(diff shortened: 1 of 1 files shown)


valldrac:
Thanks for the PR. Since we don't have visibility from the infrastructure side, we need to be cautious with this change. Did you ever hear back from the Signal team, on the forum or by email?

UjuiUjuMandan:
> Did you ever hear back from the Signal team, on the forum or by email?

Well, I suppose that's just no need for that. Neither Reddit nor Pinterest is any of their infrastructure, Google is neither. That's just how domain fronting works, utilizing another innocent domain at the same CDN provider.

> we need to be cautious with this change.

You can enable the censorship circumvention manually and see if it works, I've already confirmed it when I submitted the initial pr.

You can also confirm it's no problem by curl,

```bash
curl https://github.githubassets.com -v -H "Host: chat-signal.global.ssl.fastly.net"
```

And replace the `github.githubassets.com` with `www.gov.uk`, `www.python.org` or even `www.fastly.com`.

A HTTP 404 and a JSON response is from Signal's original server, it also has `x-signal-timestamp: xxxxxxxx` in the response header.

UjuiUjuMandan:
The only possible problem is that the 3-rd party fronting domain owner might change its CDN provider, for example, Tor Project once used `cdn.sstatic.net`, a static files domain for Stack Overflow family websites, which at some day changed to Cloudflare, and they have to choose another domain, eventually `foursquare.com`.

* https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40068
* https://github.com/net4people/bbs/issues/309#issuecomment-1974116016

That's also a SNS site what I don't recommend.

But from https://dnshistory.org, the UK government has been used Fastly for 10 years.

* https://dnshistory.org/historical-dns-records/cname/www.gov.uk

UjuiUjuMandan:
And [by the history of this file](https://github.com/signalapp/Signal-Android/commits/main/app/src/main/java/org/thoughtcrime/securesms/push/SignalServiceNetworkAccess.kt), the switching of fronting domain is quite casual, there's even no reason nor explanation in the commit message. I can't be more unwise than them.

* https://github.com/signalapp/Signal-Android/commit/db2dc473e07d507933d71bc0f7354b9b24bf68ab
* https://github.com/signalapp/Signal-Android/commit/b9474a75a43e192a5c3fee6fb87b038c306b833e
* https://github.com/signalapp/Signal-Android/commit/72777334329027c1fbca7c757489b83227a4b8fd

UjuiUjuMandan:
Closed due to inactivity.

Issue #487, opened 2025-04-26
duress password


m6n50u7:
### Guidelines

- [x] Yes, I have searched the existing requests

### Feature Description

it doesn't matter how much is signal/molly encryption strength against 5$ wrench attack. so, add the ability to chose a duress password that would delete all/specific chats and maybe change the username and pfp.

Fagner-lourenco:
Idem

testuser4433:
Very good idea

oneguynick:
Great idea

stellar-security-os:
Hi,

We’ve started working on this at Stellar Security. Before diving into the duress mechanism, we’re taking a look at how Molly stores data—especially things like attachments (images, files, etc.).

For now, it makes sense to only allow the duress password if the user has database encryption enabled. 

We’ll send over a pull request once our solution has been tested and works. We will keep you updated.

All the best.

stellar-security-os:
We've finished implementing the new duress password feature, and it's now in the testing phase.

Here’s how it works:

If a user has database encryption turned on, they’ll now see an option to enable a duress password.

If someone enters this duress password on the lock screen, all of Molly’s data will be instantly wiped from the device. The user will also be automatically signed out of the app.

We’ll be sending a pull request soon.

valldrac:
Sounds great, thanks for the update. Looking forward to the PR.

account75783:
Sounds amazing! I'd love to see this in a final version.

PR #547, opened 2025-07-25
feat: Duress code impl

stellar-security-os:
Duress code implementation based on the feature request: https://github.com/mollyim/mollyim-android/issues/487

----------

Here’s how it works:

If a user has database encryption turned on, they’ll now see an option to enable a duress password.

If someone enters this duress password on the lock screen, all of Molly’s data will be instantly wiped from the device. The user will also be automatically signed out of the app.

1_db_off
2_db_on
3_db_on_duress_on

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/ChangePassphraseDialogFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/ChangePassphraseDialogFragment.java
@@ -44,6 +44,7 @@ public class ChangePassphraseDialogFragment extends DialogFragment {
   public static final int MODE_CHANGE  = 0;
   public static final int MODE_ENABLE  = 1;
   public static final int MODE_DISABLE = 2;
+  public static final int MODE_DURESS_SET = 3;
 
   private int mode = MODE_CHANGE;
 
@@ -90,6 +91,7 @@ public class ChangePassphraseDialogFragment extends DialogFragment {
     }
 
     if (mode == MODE_ENABLE || mode == MODE_CHANGE) {
+          // || mode == MODE_DURESS_SET) { may be use validator for duress later
       initializePassphraseValidator();
     }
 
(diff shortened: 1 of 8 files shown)


valldrac:
Thanks for the PR! Really appreciate the work.

That said, this is a particularly sensitive feature. In some jurisdictions, using a duress code could legally be considered destruction of evidence, which might put users at risk. From Molly side, I think we should take extra care to reduce that risk as much as possible.

Some design principles to consider:

1. After triggering the duress code, the app should look like a clean install where the user set a passphrase but never registered. Nothing should visibly hint that the duress code was triggered.
2. If someone dumped the device storage before using the duress code, the app data should stay undecryptable even if the real password is provided later.
3. It shouldn't be possible to tell whether a duress code is configured just by inspecting the device.

Right now, the implementation has some clear issues:

After duress trigger, encryption passphrase get disabled, making the wipe evident.

It's possible to detect whether a duress code is set by measuring how long a wrong password takes. It doubles the check delay (KDF timing is ~6s vs. the usual ~3s).

The HMAC key in the Keystore is not wiped after triggering the duress code.

We should probably rethink the threat model here and iterate toward a version that offers real plausible deniability. Happy to help think through next steps.

stellar-security-os:
Thanks for the response and the quick review.

Our initial goal was to provide an instant wipe mechanism for Molly users, which worked well in the PR.

That said, your points make sense, we will do some changes where we re-think the threat-model, so the mechanism works in a better way.

We’ll work on a new PR that solves the issues you mentioned.

m6n50u7:
you also need to make sure that after the wipe, the duress passphrase becomes the new passphrase.

m6n50u7:
as for the "unlock time" difference. you can solve the issue by showing a fake "app unlocked" screen while in the background it will still be working on database wiping.

stellar-security-os:
We're discussing a solution directly with Molly. There is some cons- and pros on the solutions and ideas provided here.

We'll keep you updated.

Issue #640, opened 2025-12-02
Unify Molly and Molly-FOSS into a single FOSS app


valldrac:
Consolidate the Molly and Molly-FOSS variants into a single, fully FOSS-compliant app by replacing all proprietary components. Currently, the non-free variant is maintained to get FCM, but this can be achieved without a proprietary component. The other non-free components can't be used with Molly because of other constraints.

A single FOSS build will give all users access to a fully open-source app and reduce confusion and support requests. Accrescent users will get FCM support (fixes #516).

This change eliminates the need to maintain two separate builds and simplifies distribution.

Tasks:

- [ ] #637
- [ ] Replace GMaps with OSM
- [ ] Remove Play Store subscription and billing components
   (Cloud Backup subscriptions already don't work and cannot be fixed until Signal provides an alternative payment method, since Play Store subscriptions cannot be managed from a 3rd-party app)
- [ ] Remove conditional code paths related to FOSS vs non-FOSS variants
- [ ] Remove FCM variant and refactor CI workflows
- [ ] Link non-FOSS F-Droid repository to the FOSS repo
- [ ] Add Gradle plugin to enforce dependency license policy
- [ ] Update website and documentation

MillionsToOne:
There should be a disclaimer about the GMS thing as GMS is still proprietary.

MillionsToOne:
> There should be a disclaimer about the GMS thing as GMS is still proprietary.

I mean firebase cloud messaging, which relies on GMS and is proprietary

secretmango:
The backend yes, but the Mastodon Android app is the first one to use FCM without any proprietary libraries.

kasia-de:
If this is merged, and I used Molly on a device that supports FCM and has Google Play services, would I be able to set Molly up as a new installation, without linking it to Google via FCM IDs?

valldrac:
> If this is merged, and I used Molly on a device that supports FCM and has Google Play services, would I be able to set Molly up as a new installation, without linking it to Google via FCM IDs?

In the app settings, under Notifications, you will be able to choose between FCM, Websocket, or UnifiedPush (the latter only on main devices).

Not sure if that answers your question.

kasia-de:
> > If this is merged, and I used Molly on a device that supports FCM and has Google Play services, would I be able to set Molly up as a new installation, without linking it to Google via FCM IDs?
> 
[...]

PR #683, opened 2026-01-24
remove mention of Molly-FOSS

davidlime482:


jermanuts:
I think a note about why molly-foss no longer present in comparison chart would be useful? @valldrac

Revan335:
Maybe this is Part of the Unifing/Documentation the App: https://github.com/mollyim/mollyim-android/issues/640

Issue #289, opened 2024-02-27
Molly as linked device is sending content-less messages in my conversations


brendanny:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I have Molly as a secondary device on my Pixel tablet and after about a week, I opened Molly again on it. As part of the syncing and receiving cached messages the following happened:

1) Content-less messages appeared in my conversations in Signal on my primary device and in Molly, but not in my linked Signal Desktop
2) I assume because of these content-less messages being sent, a pending message request was accepted
3) My primary device Signal received notifications for every single message reaction that I'd received since the last time Molly synced.

### Steps to reproduce

1. Link Molly as a secondary device
2. start sending messages and using Molly/Signal

### Molly version

v7.0.1

### Android version

Android 14

### Device

Google Pixel Tablet

### Link to debug log

https://debuglogs.org/android/7.0.1/f34f238bd2a4dd4fa59a95d205acb3bfa92579de94c809350def7b34815ce9ca

valldrac:
This should be resolved in the latest versions. Let us know if it still happens.

tennisballer:
Hi @valldrac,

I'm experiencing the same "content‑less messages" bug described in this issue on the latest Molly-FOSS version [insert your version number, e.g., v8.7.3-2. I'm using GrapheneOS on a Pixel 8 Pro (primary) and two to three linked Android devices (a Pixel 10 XL and another phone/tablet), all kept online simultaneously.

The phantom messages are empty "Media message" notifications from my own number (Note to Self). They appear primarily on the linked devices, but occasionally on the primary when it comes online after being offline. Tapping them opens an empty conversation, and the ghost actually inserts itself as a real message and immediately marks itself as read—corrupting the unread state of the chat.

**What I've tested:**
- Clean installs of Molly-FOSS, with no restored backups
- Both WebSocket and UnifiedPush notification methods (ghost occurs with both)
- Reduced to only one linked device for a period—ghost still appeared, but seems more frequent with 2–3 linked devices
- Database cleanup, disabling battery optimization, force‑stopping, reboots
- Confirmed that linking a desktop (Signal Desktop) does **not** trigger the ghost—only Android linked devices do
- No VPN, firewall, or custom DNS active during testing
- I'm currently running all linked devices online simultaneously (2–3 devices), which may be exacerbating the sync chatter

**What I haven't been able to do yet:**
[...]

PR #706, opened 2026-03-04
Enable MollySocket provisioning and UnifiedPush for linked devices

### LINKED ISSUE #404: Add UnifiedPush Support for Linked Devices

### Feature description

With Molly v7.26.1-1 and NTFY 1.16.0, my second linked phone does not even show the UnifiedPush -option in notifications. WebSocket is the only option visible.

I have tried reinstalling Molly (and NTFY) two times, and no luck so far with this phone. Main phone works just fine and I have MollySocket working.

signal-2024-12-04-112200

danielsoneg:
Fixes #404 

Linked devices cannot currently use UnifiedPush with MollySocket because this requires registering a separate set of device credentials.

This MR creates an alternate flow for this case which rides off the existing behavior for registering a linked device, re-using the flow for generating a QR code for the main device to scan to create and send the new device credentials to MollySocket.

I did my best to keep this to a relatively minimal changeset so it's easier to review - I'm re-using as much of the existing behavior around the linked device registration as I can. There's a couple areas where it'd probably benefit from some additional refactoring or abstraction, but I wanted to keep this clean enough that the change and idea itself was obvious enough to evaluate on its merits.

Changes:
1. Pull out the linked device QR code provisioning UI so that it can be re-used for MollySocket provisioning
2. Create a LinkProvisioningQrContract type to expose the link provisioning status with at least a bit of abstraction so it'll be obvious if it breaks later
3. Add in a flow for provisioning another linked device for MollySocket if the current device is a linked device, wire this up across the app.
4. Remove gates around using UnifiedPush while on a linked device

(This is a new codebase and environment for me, so I used LLM assistance quite a bit, but I've reviewed both the code and the design thoroughly. Also, I obviously didn't use the LLM to write this message, nor will you have to deal with lobster-speak in the comments.)


app/src/main/java/org/thoughtcrime/securesms/registration/ui/link/RegisterLinkDeviceQrFragment.kt
danielsoneg: All of the changes in this file are from abstracting out the LinkProvisioningQrSection above

app/src/main/java/org/thoughtcrime/securesms/registration/ui/link/LinkProvisioningQrSection.kt
danielsoneg: This is a direct lift of the visual blocks from RegisterLinkDeviceQrFragment below

danielsoneg:
Thanks for letting the build run - I was struggling to get the build running locally (well, not "struggling" so much as "it takes an hour and turns my laptop into a space heater," which makes the dev cycle a bit of a challenge). Build succeeds on my machine now.

p1gp1g:
Hey ! Thanks for the contribution.

If I understand correctly, you creates a new MollySocket linked device for the linked Molly (so a common setup would be 1. Main Device, 2. Linked device, 3. MollySocket for main device, 4. MollySocket for linked device). This is interesting, as it avoids modifying MollySocket - the solutions I had in mind required some changes on MollySocket. What I'm not sure is: what is the workflow for a linked device to get UnifiedPush, can you share the steps?

To be transparent, I plan to work on web push support for the Signal server in a few weeks. If that's accepted upstream, we won't have to deal with MollySocket anymore. And it would be a lot cleaner. So, if we can go that way, it will be preferred. If we can't go with an official support, then we will consider using MollySocket for linked device as you suggest here. So, if you want we can continue to gently review it, and talk about the UX too, but that's not sure it will be used.

danielsoneg:
Hah, yeah, I just want something that works - I'm fine if this code doesn't make it in.

You're reading the flow right - although my use case here is somewhat specific: I've got an iPhone which is primary on my signal account (and runs Signal), and then I've got an android running Molly I'd like to use UnifiedPush with. I specifically avoided modifying MollySocket, although I suspect your case (two Molly devices) implies two MollySockets.

The user flow is:
1. Linked Device: settings -> Notifications -> push notifiactions -> delivery service, select UnifiedPush
2. Linked Device: code path forks on "isLinkedDevice" - goes to new provisioning flow
3. Linked Device: shows "new linked device" QR code
4. Primary Device: scan device QR code, approve
5. Linked Device: catches verification code, passes to normal linked device flow to get a device ID from Signal server
6. Linked Device: now has device ID and password, resumes normal MollySocket configuration flow (which, honestly, I've never seen, because I've only got this linked Molly device 😄)

So the fork is that instead of the device the user is currently on (the linked device) running the full flow to register a new device, we use the linked device QR code registration machinery to offload that part over to the primary device to get the Signal verification code and then use that to jump back into the standard setup flow on the linked device.

jangorecki:
@p1gp1g any update on signal server work? Battery drain on carrier data makes molly not much usable for running in the background

tennisballer:
I'm experiencing ghost 'media message' sync artifacts (#289) on linked devices using WebSocket. Would UnifiedPush on linked devices potentially help with this? I'd be happy to test a build if there's a pre‑release APK available

PaperTobi:
Any update?

p1gp1g:
The server-side implementation is done, waiting for review: https://github.com/mollyim/flatline-whisper-service/pull/20

I'll open a PR on signal once reviewed for flatline

danielsoneg:
@p1gp1g Out of curiosity, what's the relationship between flatline and signal itself? Looks like flatline is to run a completely separate signal "network", is that correct? Is there message interchange between flatline and signal?

(I swear I googled for all of this and read some stuff, just trying to check my understanding here)

Issue #303, opened 2024-03-27
Crashes without WebView


shuvashish76:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I've disabled my system webview as the updated version takes huge amount of space of low end device. I assume it requires _only_ to solve CAPTCHA? Is it possible to redirect this to users' browser & solve them from there only?

<details><summary>Logs:</summary>
<p>

```
FATAL EXCEPTION: main
Process: im.molly.app, PID: 1325
android.view.InflateException: Binary XML file line #41: Binary XML file line #41: Error inflating class android.webkit.WebView
Caused by: android.view.InflateException: Binary XML file line #41: Error inflating class android.webkit.WebView
Caused by: java.lang.reflect.InvocationTargetException
	at java.lang.reflect.Constructor.newInstance0(Native Method)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
	at android.view.LayoutInflater.createView(LayoutInflater.java:647)
	at com.android.internal.policy.PhoneLayoutInflater.onCreateView(PhoneLayoutInflater.java:58)
	at android.view.LayoutInflater.onCreateView(LayoutInflater.java:720)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:788)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730)
	at android.view.LayoutInflater.rInflate(LayoutInflater.java:863)
	at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:824)
	at android.view.LayoutInflater.rInflate(LayoutInflater.java:866)
	at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:824)
	at android.view.LayoutInflater.inflate(LayoutInflater.java:515)
	at android.view.LayoutInflater.inflate(LayoutInflater.java:423)
	at org.thoughtcrime.securesms.registration.fragments.CaptchaFragment.onCreateView(CaptchaFragment.java:35)
	at androidx.fragment.app.Fragment.performCreateView(Fragment.java:3114)
	at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:557)
	at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:272)
	at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1943)
[...]

PR #736, opened 2026-04-22
fix(captcha): fall back to external browser when system WebView is disabled (refs #303)

jim-daf:
### Problem

Issue #303 reports that disabling / removing the system
WebView (a common configuration on low-storage devices and
some hardened ROMs) bricks registration: `CaptchaFragment`
unconditionally calls into the WebView and either silently
fails to render or crashes when the WebView constructor
throws `AndroidRuntimeException("WebView ... not available")`.

### Fix

Wrap the WebView access in a try/catch. If accessing the
WebView (`binding.registrationCaptchaWebView` and its
`settings`) throws, or if `loadUrl` throws, fall back to
launching `BuildConfig.SIGNAL_CAPTCHA_URL` via
`Intent.ACTION_VIEW`. The existing `signalcaptcha://` URI
scheme handler (the same one the share-sheet uses today)
brings the user back into the app with the token and
`handleCaptchaToken` runs as normal.

If no browser is available either we just `navigateUp()` so
the user can back out and retry, no worse than the current
blank-screen behaviour.

### Why this is in scope of `mollyim-android`

Upstream Signal does not have to care because it does not
ship a browser-light user base. The mollyim user base has
explicitly asked for it (#303) and the change is
self-contained inside the fork's existing CaptchaFragment
(no AndroidManifest or shared-component edits).

### Notes

* The intent filter for `signalcaptcha://` is already
  declared (it is what makes the share-sheet workflow
  function today), so the round-trip through the browser
  works without any manifest changes.
* `clearCache(true)` is preserved on the WebView path; the
  browser path does not need it because the user's browser
  manages its own cache.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/registration/ui/captcha/CaptchaFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/registration/ui/captcha/CaptchaFragment.kt
@@ -6,6 +6,9 @@
 package org.thoughtcrime.securesms.registration.ui.captcha
 
 import android.annotation.SuppressLint
+import android.content.ActivityNotFoundException
+import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
 import android.view.View
 import android.webkit.WebView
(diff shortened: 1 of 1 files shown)


valldrac:
Thanks for submitting this, but it's fundamentally flawed in its understanding of Android development and does not provide the functionality it claims. It also seems to have been vibe coded. Please take the time to deeply understand the changes and carefully test before submitting future contributions.

PR #742, opened 2026-04-28
Fix failure loop when sending messages after push method update

### LINKED ISSUE #732: Unable to send messages (single, spinning circle)
### Guidelines

- [x] Yes, I have searched for existing issues

### Bug Description

Hello,

since this morning I am unable to send messages.
The message indicator is a single spinning circle.

I am  running Molly v.8.7.2-1 from Accrescent store on latest GrapheneOS.

Interestingly, I am able to receive messages without issue.

For push notifications I use MollySocket.

I already rebooted the device without success.

I wanted to transmit the debug log but I got an error that transmission is not possible, that is why I attached it here.

[molly-log-1776436684755.zip](https://github.com/user-attachments/files/26829712/molly-log-1776436684755.zip)

Best regards.

### Screenshots

_No response_

### App Version

v.8.7.2-1

### Device Model

Pixel 10

### Android Version

Android 16

### Link to Debug Log

_No response_

p1gp1g:
Fix #732

There were actually 2 bugs:
- When deleting the linked device, we didn't update isMultiDevice, which led to sending sync requests to an empty list of devices, causing an error loop
- When linking to a new MollySocker linked device, we didn't add any pre-keys, which led to this device being [filter out, as an inactive device](https://github.com/mollyim/mollyim-android/blob/914315b220916b50cecd1ecf1a8daaeaf0a9e357/lib/libsignal-service/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L2830), causing sending sync requests to an empty list of devices, causing the same error loop

I don't know exactly why we didn't have that bug earlier, I suppose this is a new mechanism to filter out messages to stalled devices

Additionally, something should be done to prevent sending to an empty list of devices (this can happen only when sending messages to ourselves), causing this error 400. We can either:
1. When the [deviceIds](https://github.com/mollyim/mollyim-android/blob/914315b220916b50cecd1ecf1a8daaeaf0a9e357/lib/libsignal-service/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L2830) after filter is empty, we add a dummy deviceId `2` => so we can get a 409, and Extra/Missing devices will be returned by the server
2. Or return without processing the different `sendMessage` function when `OutgoingPushMessage.messages` is empty
3. Update SignalServer to return a 409 instead of 400 when `messages` is empty

I think the cleaner would be the 3rd option, but we don't have the hand on it. Then, even if it sounds more hacky, I'd go with the first solution, so any new function relying on getEncryptedMessage wouldn't reintroduce the bug.

### DIFF
--- a/app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
+++ b/app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
@@ -94,7 +94,7 @@ object MollySocketRepository {
 
   @Throws(IOException::class)
   fun removeDevice(device: MollySocketDevice) {
-    AppDependencies.linkDeviceApi.removeDevice(device.deviceId).successOrThrow()
+    LinkDeviceRepository.removeDevice(device.deviceId)
   }
 
   fun getDeviceStatus(device: MollySocketDevice): LinkStatus {
--- a/app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
+++ b/app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
@@ -15,17 +15,25 @@ import okhttp3.RequestBody.Companion.toRequestBody
 import org.signal.core.util.Base64
 import org.signal.core.util.Util
 import org.signal.core.util.logging.Log
+import org.signal.libsignal.protocol.SessionBuilder
+import org.signal.libsignal.protocol.SignalProtocolAddress
 import org.signal.libsignal.protocol.util.KeyHelper
 import org.thoughtcrime.securesms.AppCapabilities
+import org.thoughtcrime.securesms.crypto.ReentrantSessionLock
 import org.thoughtcrime.securesms.dependencies.AppDependencies
+import org.thoughtcrime.securesms.dependencies.AppDependencies.keysApi
 import org.thoughtcrime.securesms.keyvalue.PhoneNumberPrivacyValues.PhoneNumberDiscoverabilityMode
 import org.thoughtcrime.securesms.keyvalue.SignalStore
 import org.thoughtcrime.securesms.linkdevice.LinkDeviceRepository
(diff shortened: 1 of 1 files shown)


app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
```diff
@@ -87,14 +96,37 @@ object MollySocketRepository {
...
+        SignalProtocolAddress(recipient.identifier, deviceId)
+      )
+    )
+    sessionBuilder.process(preKey)
+  }
+
   @Throws(IOException::class)
   fun removeDevice(device: MollySocketDevice) {
-    AppDependencies.linkDeviceApi.removeDevice(device.deviceId).successOrThrow()
+    LinkDeviceRepository.removeDevice(device.deviceId)
```
valldrac, line 128: It should `throw IOException("removeDevice failed")` if `LinkDeviceRepository.removeDevice` return false.

valldrac (changes requested):

app/src/main/java/im/molly/unifiedpush/MollySocketRepository.kt
```diff
@@ -87,14 +96,39 @@ object MollySocketRepository {
       accountAttributes,
       aciPreKeyCollection, pniPreKeyCollection,
       null
-    ).also {
+    ).also { deviceId ->
       SignalStore.account.isMultiDevice = true
+      loadPreKeys(deviceId)
```
valldrac: `loadPreKeys` makes a network request, so set `isMultiDevice` only after it succeeds, just in case:

```
loadPreKeys(deviceId)
SignalStore.account.isMultiDevice = true
```

valldrac (changes requested):

PR #747, opened 2026-05-04
feat(kids flavor): opt-in parental controls for family Signal use

avacar:
## Summary

This PR adds an **opt-in `kids` Gradle build flavor** to Molly that enables parental controls
for families who want to use Signal-compatible messaging with young children.

All changes are behind the `kids` flavor flag. The existing `prod` and `staging` flavors are
completely unaffected — this is purely additive.

---

## Motivation

Families with young children often want a secure, private messaging channel with grandparents,
relatives, and family friends. Signal is an excellent technical foundation for this, but it has
no concept of a "restricted" child account. The child can message anyone, see all contacts, and
change account settings.

This fork adds a PIN-gated parental layer so parents can:
- Allow only specific group chats to be visible to the child
- Block the child from starting new conversations or making calls
- Prevent access to account settings, Stories, and linked device management
- Manage the allow-list from a parent settings panel on the child's device

Registration uses free Google Voice numbers so children never expose personal phone numbers.
Communication happens via Signal usernames, requiring no app changes.

---

## What's in this PR

| Phase | Feature |
|---|---|
| 0 | `kids` Gradle build flavor + `parentalModeEnabled` flag |
| 1 | `ParentalControlValues.kt` — PIN storage (SHA-256 + random salt) and allowed-thread list |
| 2 | Conversation list filtering — show only allowed threads |
| 3 & 4 | Block new outbound conversations/calls; gate group invites behind PIN |
| 5 | Suppress notifications from non-allowed threads |
| 6 | Hide Stories tab and lock account/linked-device settings |
| 7 | `ParentalControlActivity` — PIN-gated parent settings panel on the child's device |

**62 unit tests** cover the parental-control layer. All pass on the current build.

---

## Architecture — why this is safe to merge

**New files only for parental logic.** All parental UI and data live in:
- `app/src/main/java/.../keyvalue/ParentalControlValues.kt`
- `app/src/main/java/.../parental/` (Activity, ViewModel, dialogs)

**Minimal touch of existing files.** Guard clauses are added at a handful of call sites:
```kotlin
if (SignalStore.parental().parentalModeEnabled) { /* restrict */ }
```
No Signal protocol, crypto, database schema, or network code is modified.

**Zero-conflict upstream tracking.** This branch is based on Molly's current `main` and has
been kept in sync through multiple upstream releases with zero merge conflicts.

**Opt-in at build time.** A standard `./gradlew assembleProdRelease` produces the normal Molly
APK. The parental controls only activate in a `kidsRelease` build.

---

## Public fork

The full fork with history and documentation is at:
**https://github.com/avacar/mollykids-android**

---

## Notes for reviewers

- If this scope is too large to consider for mainline Molly, I'm happy to discuss a narrower
  contribution (e.g. just the `kids` flavor scaffolding in Phase 0, with the rest maintained
  downstream).
- The `ParentalControlActivity` is intentionally a standalone `AppCompatActivity` rather than
  a destination in `AppSettingsActivity`'s NavGraph, to avoid upstream merge conflicts in the
  settings navigation graph.
- The PIN is stored as `SHA-256(salt + PIN)` with a random 16-byte salt. No plaintext PIN is
  ever persisted.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,5 @@
 org.gradle.jvmargs=-Xmx12g -XX:+UseParallelGC
+org.gradle.java.home=C:/Program Files/Eclipse Adoptium/jdk-21.0.10.7-hotspot
 android.useAndroidX=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.problems=fail
@@ -10,6 +11,7 @@ org.gradle.parallel=true
 # this flag adds an extra layer of protection.
 # See: https://docs.gradle.org/current/userguide/toolchains.html#sub:disable_auto_provision
 org.gradle.java.installations.auto-download=false
+org.gradle.java.installations.paths=C:/Program Files/Eclipse Adoptium/jdk-21.0.10.7-hotspot
 
 # Test fixtures support for Android modules
 android.experimental.enableTestFixturesKotlinSupport=true
--- a/README-MOLLYKIDS.md
+++ b/README-MOLLYKIDS.md
@@ -1,0 +1,116 @@
+# MollyKids
+
+**MollyKids** is a downstream fork of [Molly](https://github.com/mollyim/mollyim-android) — a
+security-hardened Signal client — that adds parental controls for family use. It lets children
+communicate with pre-approved family members via Signal-compatible messaging on semi-locked-down
+tablets, while keeping parents in full control of who the child can reach.
+
+---
+
+## Why MollyKids?
+
+Signal is excellent for secure family communication, but it has no concept of a restricted
+child account. MollyKids adds a PIN-gated parental layer that:
+
+- Shows children **only the group chats parents explicitly allow** — no contacts list, no new
+  conversations
+- **Blocks outbound messaging** to any thread not on the allow-list
+- **Suppresses notifications** from non-allowed threads so nothing leaks through
+- **Hides Stories, account settings, and linked-device management** from the child
+- Lets parents manage the allow-list and change the PIN via a PIN-gated Settings panel on the
+  device itself
+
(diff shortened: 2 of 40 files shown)


bluecat21:
This is great!

I happen to be aware of parents who will not let their children use anything other than a landline or dumb phone without images if that exists because they are concerned about their child seeing genitalia, gore, etc. from their school friends who don't have a locked down device. 

I propose these remedies: allow disabling receiving images and videos, video calls, and eventually use an AI model to detect nudity and gore and block it. This is provided as part of AICore (app included on Google Mobile Services devices) and does function locally, but an open source alternative could be provided in the future and since it will likely use significant storage it should be provided as an addon pack via another app (in the same way as icon packs) which can be updated by Molly to avoid dynamic code loading or wasting storage for users who don't want this.

Separately, allow disabling sending images and videos to prevent them from sending inappropriate images. 

It probably makes sense for the settings for sending/receiving  images/videos and video calls to be per-contact.

bluecat21:
And possibly somewhat controversially, deleted messages should be able to be sent to a quarantine so that parents can review them later.

I know this might upset some people but I dare deeply for children and this would not impact the privacy, security, freedom or anything else of adults and would allow concerned parents to choose a safe option where if their children make a mistake, their messages are not stored in a central database, contrary to mainstream "kid friendly" messengers or SMS monitoring tools.

gradle.properties
kuflierl, line 1: You do not want to hard-code paths in the build system that will break everything other than windows or for a matter of fact, every build that doesn't use Eclipse Adoptium 21.0.10.7-hotspot

avacar: good point.  I'm addressing it.

PR #778, opened 2026-07-03
Fix linking a device: advertise usernameChangeSyncMessage capability (Fixes #770)

### LINKED ISSUE #770: Linking device doesn't work anymore
### Guidelines

- [x] Yes, I have searched for existing issues

### Bug Description

When I want to link a new device from my main phone (Pixel 10 GraphenOS) with Signal on my other device, the linking process didn't stop to loop.

I already tried to remove app, reinstall Molly, clean cache.. but that's doesn't change anything. I don't understand why on my second phone with Molly the process to link device don't work.

### Screenshots

Image

### App Version

v8.7.3-2

### Device Model

Poco F6 Pro

### Android Version

LineageOS Android 16

### Link to Debug Log

_No response_

timdavidfriedrich:
## Problem
Linking Molly as a secondary device to a current Signal primary fails. The QR is scanned but the device never registers and the screen spins indefinitely. Fixes #770.

## Root cause
`PUT /v1/devices/link` is rejected with HTTP 409. The server's capability-downgrade check requires a new linked device to advertise every `preventDowngrade` capability the account already has. Current Signal clients advertise `usernameChangeSyncMessage` (added upstream, `preventDowngrade=true`), so the account has it, but Molly doesn't send it, so the link is treated as a downgrade and refused.

## Fix
Advertise `usernameChangeSyncMessage` in the device capabilities, matching current Signal. Added to `AccountAttributes.Capabilities` (defaulted, so other call sites are untouched) and set to true in `AppCapabilities`.

A second commit resets `isRegistering` on link failure so an error dialog is shown instead of an indefinite spinner. Previously any post-scan failure was masked by the progress dialog, which is why this surfaced as an endless loop.

## Testing
Built `prodWebsiteDebug` and linked against a primary running Signal 8.17. 

* **Before:** HTTP 409, endless spinner

* **After:** Link succeeds (server assigns a device id) and the device appears in the primary's linked-devices list. No dependency/native changes.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/AppCapabilities.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/AppCapabilities.kt
@@ -13,7 +13,8 @@ object AppCapabilities {
       storage = storageCapable,
       versionedExpirationTimer = true,
       attachmentBackfill = true,
-      spqr = true
+      spqr = true,
+      usernameChangeSyncMessage = true
     )
   }
 }
--- a/app/src/main/java/org/thoughtcrime/securesms/registration/ui/link/RegisterLinkDeviceQrViewModel.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/registration/ui/link/RegisterLinkDeviceQrViewModel.kt
@@ -146,6 +146,7 @@ class RegisterLinkDeviceQrViewModel : ViewModel() {
   fun clearErrors() {
     store.update {
       it.copy(
+        isRegistering = false,
         showProvisioningError = false,
         registrationErrorResult = null
       )
@@ -156,7 +157,10 @@ class RegisterLinkDeviceQrViewModel : ViewModel() {
 
   fun setRegisterAsLinkedDeviceError(result: RegisterLinkDeviceResult) {
     store.update {
-      it.copy(registrationErrorResult = result)
+      it.copy(
+        isRegistering = false,
+        registrationErrorResult = result
+      )
     }
   }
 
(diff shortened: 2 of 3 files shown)


p1gp1g:
This is indeed the fix to apply, because the server has introduced USERNAME_CHANGE_SYNC_MESSAGE with preventDowngrade=true https://github.com/signalapp/Signal-Server/commit/81167a8ae59decdd20a98b0319cefbdf05b298a1

But it is already included in the WIP new version of Molly, from upstream: https://github.com/mollyim/mollyim-android/blob/molly-8.18.1/app/src/main/java/org/thoughtcrime/securesms/AppCapabilities.kt

At least, we know it will be fixed in the next release, so we don't have to dig into this issue, thanks for your help :+1:

timdavidfriedrich:
> This is indeed the fix to apply, because the server has introduced USERNAME_CHANGE_SYNC_MESSAGE with preventDowngrade=true [signalapp/Signal-Server@81167a8](https://github.com/signalapp/Signal-Server/commit/81167a8ae59decdd20a98b0319cefbdf05b298a1)
> 
> But it is already included in the WIP new version of Molly, from upstream: https://github.com/mollyim/mollyim-android/blob/molly-8.18.1/app/src/main/java/org/thoughtcrime/securesms/AppCapabilities.kt
> 
> At least, we know it will be fixed in the next release, so we don't have to dig into this issue, thanks for your help 👍


Alright, sounds good, thanks. I'll close the issue, then.

timdavidfriedrich:
@p1gp1g Although, the second fix might be worth implementing (https://github.com/mollyim/mollyim-android/pull/778/changes/dde0b3d0af5c35a1de1c1dea101d0c2085c39376):
> A second commit resets `isRegistering` on link failure so an error dialog is shown instead of an indefinite spinner. Previously any post-scan failure was masked by the progress dialog, which is why this surfaced as an endless loop.

This might be an UX improvement in case of future issues. The small change prevents the loading spinner from spinning endlessly, which could be confusing for some users.

p1gp1g:
Yes, definitely. 8.19.1 is being merged: do you prefer opening a PR on this coming branch yourself, or do you want me to open a PR for you, once the branch is pushed ?

timdavidfriedrich:
No worries, I can open a new PR myself once the new branch is pushed

PR #795, opened 2026-07-19
Add JVM args to improve R8 determinism for reproducibility

BarbossHack:
Actually, Molly is no longer reproducible (see this workflow: [29643368147](https://github.com/BarbossHack/reproducible/actions/runs/29643368147/job/88077412428))

> image

.

I first reported the issue upstream in https://github.com/signalapp/Signal-Android/issues/14809 and it has already been fixed. However, it seems that not all of the fixes have been merged into Molly.

The relevant code was taken from an upstream commit made two months ago: https://github.com/signalapp/Signal-Android/commit/09822d3ae9879565f78fd8a62e4393599df929e9

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,11 @@
-org.gradle.jvmargs=-Xmx12g -XX:+UseParallelGC
+# R8 in AGP 9.x is non-deterministic, producing intermittent dex differences between
+# otherwise-identical builds. Two flags work together to stabilize it:
+#   - -Dcom.android.tools.r8.threadCount=1 removes thread-scheduling-dependent ordering.
+#   - -XX:hashCode=3 (deterministic incrementing Object.identityHashCode) removes
+#     IdentityHashMap iteration-order non-determinism inside R8 (e.g. in
+#     KotlinClassMetadataReader, which decides whether to keep the Signature attribute
+#     on Kotlin generic lambda subclasses).
+org.gradle.jvmargs=-Xmx12g -Xms256m -XX:MaxMetaspaceSize=1g -Dcom.android.tools.r8.threadCount=1 -XX:+UnlockExperimentalVMOptions -XX:hashCode=3
 org.gradle.parallel=true
 
 android.useAndroidX=true


valldrac:
Thank you. I'd like to review this further before accepting the changes. I found out the first option, `com.android.tools.r8.threadCount`, does not exist in R8 and thus has no effect.

BarbossHack:
Indeed... I tried setting the thread count to an invalid value (-1), and it seems to have no effect (and nothing in logs), and cannot find it in r8 source code

A (dirty?) workaround would be to set `-Dcom.android.tools.r8.deterministicdebugging=true`

https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/utils/InternalOptions.java#164
```java
  // Set to true to run compilation in a single thread and without randomly shuffling the input.
  // This makes life easier when running R8 in a debugger.
  public static final boolean DETERMINISTIC_DEBUGGING =
      System.getProperty("com.android.tools.r8.deterministicdebugging") != null;
  ```

valldrac:
> A (dirty?) workaround would be to set `-Dcom.android.tools.r8.deterministicdebugging=true`

This is what I've been testing. Initially, I thought that since `threadCount` had no effect, it might not be necessary to fix the bug. But I ran tests and R8 was still randomly removing the class metadata attribute (see [this commit](https://github.com/signalapp/Signal-Android/commit/09822d3ae9879565f78fd8a62e4393599df929e9#diff-9a67f823fbc4770cd18730d234ebccb771cc7c749fb26dd9ef532211aaf010fbR23) for context).

In the end, enabling `com.android.tools.r8.deterministicdebugging` resolved the issue. I also confirmed this by taking heap dumps during the minify task and inspecting R8's configuration. The correct approach is indeed to enable `deterministicdebugging`.

You might want to report this to upstream. Looking at R8 code, this flag provides extra benefits for reproducible builds beyond enforcing single-threaded execution.

Applied in https://github.com/mollyim/mollyim-android/commit/df8936220f41314be4dba4e7ae142c23522d6808 instead of merging this PR, because our build differs slightly from upstream. We use `gradle.properties.docker` for reproducible builds and different JVM memory parameters.

BarbossHack:
Thank you for your feedback, I will report this upstream :pray:

